PR #10, opened 2018-05-30
Upgrade packages

OzzieOrca:
- angular.js to 1.7
- webpack to v4


OzzieOrca:
Thanks for the review. And thanks for making me revisit the `manifest` chunk. The docs seem to be missing discussion of it or seem outdated. I was confused about it too. I did see https://medium.com/webpack/webpack-4-changes-part-1-week-24-25-fd4d77674e55#dbe8 which makes it sound like it isn't needed anymore. I tested it locally just now and if I change an app file and run the build, I only get a new hash for app.js. vendor.js keeps the same hash which is good. There don't seem to be any official migration docs. https://github.com/webpack/webpack/wiki/Changelog-WIP has some changes but I don't see the manifest mentioned.

bryaneaton13:
Hmm, strange. Maybe it's just not documented other than that issue. But everything else looks good.

PR #14, opened 2018-06-20
Point development apiUrl at staging

OzzieOrca:
@adammeyer You want this? I'm not really using the localhost API anymore.


adammeyer:
yes!

PR #19, opened 2018-06-27
 MHP-1651: User Preferences

OzzieOrca:
- Show language names from translations.json language codes


OzzieOrca:
I've got to figure out what's going on with the Karma tests. Seems like babel isn't compiling `const` somewhere...

app/assets/javascripts/angular/missionhubApp.run.js
```diff
@@ -22,6 +23,11 @@ angular
             $analytics.pageTrack($window.location.pathname);
         }
 
+        $rootScope.whiteBackground = false;
+        $transitions.onSuccess({}, transition => {
+            $rootScope.whiteBackground = !!transition.to().whiteBackground;
```
adammeyer: what is this for, the white background?

PR #21, opened 2018-07-16
Dynamic public path

OzzieOrca:
Webpack docs: https://webpack.js.org/concepts/output/#advanced


app/assets/javascripts/angular/main.js
```diff
@@ -1,3 +1,8 @@
+/* eslint-disable angular/window-service */
+if (window.__webpack_public_path__) {
```
adammeyer: What is setting this?

OzzieOrca: This https://github.com/CruGlobal/missionhub/pull/1088/files#diff-b2d1ace20299175e4fe452c5d488ecb5R22. It feels hacky but that's the easiest place to grab it from right now. I could write code here to look at localStorage and then look at the current url and map it to cloudfront if you think that's better.

PR #23, opened 2018-07-17
Merge sunset-legacy branch

OzzieOrca:
Adds JS translation features and Netlify config

I'll merge this tomorrow morning and then we can work off master after that for our other PRs.

You've already approved this code in other PRs, just wanted to make you aware of what I'm doing and run it by you. I've merged it into staging and it runs fine.


app/assets/javascripts/angular/components/editAddress/editAddress.html
```diff
@@ -20,7 +20,7 @@ <h3 class="modal-title" id="modal-title">
         </label>
         <label ng-repeat="addressField in ['address1', 'address2', 'address3', 'address4', 'city']">
             <div class="row">
-                <span class="normal-column label">{{('general.' + addressField) | t}}</span>
+                <span class="normal-column label">{{('general.' + addressField || 'other') | t}}</span>
```
adammeyer: what does 'other' mean here?

adammeyer:
Do you want all the sunset legacy components in master?  Why would it be needed there?

app/assets/javascripts/angular/components/editAddress/editAddress.html
OzzieOrca: You know, idk why I did that. Looks like `addressField` should always be defined since it's got a static array up there.

I have a faint memory of there being an `Other` translation that had an empty string as the key. Maybe I was trying to fix that... I don't like it but there are a lot of cases where the translation key is dynamically generated. Something (API or a service or something) generates a string like "home" or "work" or something and it uses that to go look up a translation.

I can remove it.

OzzieOrca:
It's not needed in production (although we could load different JS in production too if we wanted to test against prod data). I don't think it'll hurt anything. That localStorage key should never get set. I could write code to explicitly do nothing when the Rails env is prod.

The Netlify code needs to be on staging for Netlify to function and the dynamic path stuff needs to be in every PR so the build output can use it. Having it on master and basing off of it seems to make sense.

You have any ideas for isolating it or cleaning it up? We should just be able to rip it out when we finally kill the Rails app.

adammeyer:
It's fine.  I was just making sure you knew it was going to master.

OzzieOrca:
Ya it's a good thing to check. https://github.com/CruGlobal/missionhub-web/pull/23/files#diff-7129a9286dc8510eac447b632e11e04fR3 should never happen since nothing should set that window var. Everything else just reads localStorage and shows the top bar. I'll leave the part that's in the missionhub repo off of production since it doesn't need to be there and exposes a way to change where the JS is loaded by query param. Could create a security issue.

PR #27, opened 2018-08-06
Org bulk contact import

adammeyer:
There is some survey stuff in here, but that is because there was a little overlap.  Mainly css.

But this is now independent and can go live before or after the survey settings PR.


app/assets/javascripts/angular/components/organizationContactImport/organizationContactImportStep1.component.js
```diff
@@ -0,0 +1,40 @@
...
+        const input = document.createElement('input');
+        input.setAttribute('type', 'file');
+        input.click();
+
+        input.addEventListener(
+            'change',
+            () => {
+                this.selectedFile = input.files[0];
+
+                if (this.selectedFile.type !== 'text/csv') {
```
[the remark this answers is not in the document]
adammeyer: No, I think just csv

app/assets/javascripts/angular/components/organizationContactImport/organizationContactImportStep1.html
```diff
@@ -4,6 +4,10 @@ <h2 class="normal f3 fw6 blue">
     </h2>
     <p class="lh-copy grey-darker f4 w-90">
         {{ 'contact_import:step_1.instructions' | t }}
+        <a ng-href="/surveys?organization_id={{$ctrl.organizationOverview.org.id}}" class="blue-light">
```
OzzieOrca: Don't you have a modal for doing this?

app/assets/javascripts/angular/services/survey.service.js
```diff
@@ -8,15 +9,35 @@ function surveyService(
...
+                        },
+                    ),
+                    $http.get(
+                        envService.read('apiUrl') +
+                            '/surveys/' +
+                            surveyId +
+                            '/questions',
+                    ),
+                ])
+                .then(questions => {
```
OzzieOrca: Could do `.then(([predefinedQuestions, surveyQuestions]) => {`

app/assets/javascripts/angular/components/organizationContactImport/organizationContactImportStep1.html
adammeyer: different PR

app/assets/javascripts/angular/services/survey.service.js
adammeyer: is that better?  I still need the `.data.included` part.

OzzieOrca: It just uses ES6 destructuring. It's syntactic sugar. I do like it better because you don't have magic numbers running around later when you try to pull data out and it's clear what the promise returns from the top of the function.

adammeyer: How would you suggest I use them in the function?
`const predefinedQuestions = predefinedQuestions.data.included;`
and
`predefinedQuestions = predefinedQuestions.data.included;` are invalid to the linter.

OzzieOrca: What does it say? You shouldn't have to manually parse the included array if you use the httpProxy. You should just be able to follow the relationships in the object...

OzzieOrca: Oh the names are the same? Ya idk. Rename them or disable the rule...

adammeyer: I need to use a variable in the middle of the url.  That's why I used the $http service.

OzzieOrca: The first param of the httpProxy is just a url. I did it here https://github.com/CruGlobal/missionhub-web/pull/28/files#diff-dc044f10265abca967251e5d82e4f42aR20. I know most places it has that confusing model service there instead but it just returns a url.

app/assets/javascripts/angular/components/organizationContactImport/organizationContactImportStep3.component.js
```diff
@@ -0,0 +1,156 @@
...
+
+            //map answers
+            _.forEach(this.columnMap, (question, questionIndex) => {
+                const answerId = uuidv1();
+
+                postData.included.push({
+                    type: 'answer',
+                    id: answerId,
+                    attributes: {
+                        question_id: Number(question.id),
```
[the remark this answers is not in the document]
adammeyer: They where numbers in the docs, but I don't know if they need to be.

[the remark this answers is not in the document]
OzzieOrca: @sheldond Are these IDs supposed to be strings or numbers?

app/assets/javascripts/angular/components/organizationContactImport/organizationContactImport.component.js
```diff
@@ -0,0 +1,34 @@
+import template from './organizationContactImport.html';
+import './organizationContactImport.scss';
+import '../surveyOverview/surveyOverview.scss';
+
+angular.module('missionhubApp').component('organizationContactImport', {
+    require: {
+        organizationOverview: '^',
```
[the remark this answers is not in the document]
OzzieOrca: Resolved. I removed the requires in sub components.

app/assets/javascripts/angular/components/organizationContactImport/organizationContactImportStep3.component.js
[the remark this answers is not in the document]
sheldond: In general I believe the API will accept either, we haven't been strict about the type on ids.

app/assets/javascripts/angular/components/organizationContactImport/organizationContactImportStep3.component.js
```diff
@@ -0,0 +1,156 @@
...
+                type: 'answer_sheet',
+                relationships: {
+                    survey: {
+                        data: {
+                            type: 'survey',
+                            id: this.selectedSurvey.id,
+                        },
+                    },
+                },
+                answers: {
```
OzzieOrca: Are these supposed to be under the `relationships` object?

adammeyer: Fixed.  Thanks

app/assets/javascripts/angular/components/organizationContactImport/organizationContactImportStep3.component.js
```diff
@@ -146,11 +146,13 @@ function organizationContactImportStep3Controller(
...
 
         this.disableButtons = true;
+        this.importErrors = [];
         surveyService.importContacts(postData).then(
             () => {
                 this.next();
             },
-            () => {
+            response => {
+                this.importErrors = response.data.errors[0].detail.bulk_data;
```
OzzieOrca: Seems like some of this may be undefined if you get a different kind of server error. Like a 500 or 503 or something.

app/assets/javascripts/angular/components/organizationContactImport/organizationContactImportStep3.component.js
```diff
@@ -0,0 +1,165 @@
...
+) {
+    this.contactLabels = {};
+    this.disableButtons = true;
+
+    this.$onInit = () => {
+        //parse csv
+        Papa.parse(this.selectedFile, {
+            complete: results => {
+                if (results.errors.length) {
+                    $log.error('CSV Parse errors', results.errors);
```
OzzieOrca: This error needs to be shown to the user if it will prevent them from getting the parsed data. Looks like they just end up with undefined csvData. We could log this and let them continue if the parse errors aren't fatal.

PR #28, opened 2018-08-06
MHP-1674: Survey Responses

OzzieOrca:


app/assets/javascripts/angular/components/organizationOverview/organizationOverview.component.js
```diff
@@ -31,7 +31,9 @@ function organizationOverviewController(
     vm.toggleVisibility = userPreferencesService.toggleOrganizationVisibility;
 
     vm.showOrgNav = () => {
-        return $state.$current.name !== 'app.ministries.ministry.survey';
+        return !$state.$current.name.startsWith(
```
adammeyer: Is the browser compatibility okay for this? `startsWith`

app/assets/javascripts/angular/components/organizationOverviewSurveys/organizationOverviewSurveys.html
```diff
@@ -56,12 +56,11 @@
...
             <div uib-dropdown>
-                <ng-md-icon uib-dropdown-toggle icon="more_vert" size="26"
-                            ng-click="$ctrl.deleteSurvey(); $event.stopPropagation();"></ng-md-icon>
+                <ng-md-icon uib-dropdown-toggle icon="more_vert" size="26"></ng-md-icon>
 
                 <div class="dropdown-menu" uib-dropdown-menu role="menu" aria-labelledby="single-button">
-                    <a class="dropdown-item" ui-sref="app.ministries.ministry.survey({ surveyId: survey.id })">{{ 'surveyTab:menu.edit' | t }}</a>
-                    <a class="dropdown-item" ng-href="/survey_responses/new?preview=true&survey_id={{survey.id}}">{{ 'surveyTab:menu.preview' | t }}</a>
+                    <a class="dropdown-item" ui-sref="app.ministries.ministry.survey.manage({ surveyId: survey.id })">{{ 'surveyTab:menu.edit' | t }}</a>
+                    <a class="dropdown-item" ng-href="/survey_responses/new?preview=true&survey_id={{survey.id}}" target="_blank">{{ 'surveyTab:menu.preview' | t }}</a>
```
adammeyer: Should this open in a new tab?  Maybe check with Alex on that.

OzzieOrca: Her answer is yes. Thanks for checking.

app/assets/javascripts/angular/components/organizationOverview/organizationOverview.component.js
OzzieOrca: I guess edge doesn't support it and babel needs core-js to support it. I'll find something else.
http://kangax.github.io/compat-table/es6/#test-String.prototype.startsWith
http://kangax.github.io/compat-table/es6/#babel-core-js-note

adammeyer: I'm surprised edge doesn't support it.  Thanks for checking

app/assets/javascripts/angular/components/peopleFiltersPanel/peopleFiltersPanel.component.js
```diff
@@ -86,44 +84,68 @@ function peopleFiltersPanelController(
...
-            assignedTos: getTruthyKeys(vm.filters.assigned_tos),
-            groups: getTruthyKeys(vm.filters.groups),
+            searchString: this.filters.searchString,
+            includeArchived: this.filters.includeArchived,
+            labels: getFiltersFromObj(this.filters.labels),
+            assignedTos: getFiltersFromObj(this.filters.assigned_tos),
+            groups: getFiltersFromObj(this.filters.groups),
+            statuses: getFiltersFromObj(this.filters.statuses),
+            genders: getFiltersFromObj(this.filters.genders),
+            questions: _.flow([
```
adammeyer: fancy

OzzieOrca: I've seen pipe in functional programming libraries like https://ramdajs.com/docs/#pipe and wanted to use it. I was considering making a pipe function with Reduce but found _.flow. If a function takes more than one param and the data passed into it is the first param it won't work.

Like this won't work cuz it calls map like `_.map(iteratorFn)` and expects it to return a function with the signature of `myMapFn(dataToMap)`. Because lodash doesn't support partial application of a function (providing only some of the params and expecting it to return a function that accepts the rest) and _.map expects the data first this won't work.
```
_.flow([
  _.map(item => item.value)
]);
```

This would work but it's messier
```
_.flow([
  data => _.map(data, item => item.value)
]);
```

That's why functional libraries like Ramda and lodash-fp have the data as the last parameter and do function currying so can pass all the params except for the data and it will give you a new function that accepts the data to process. It helps you combine function easier.

app/assets/javascripts/angular/components/peopleScreen/peopleScreen.html
```diff
@@ -13,11 +13,35 @@
     </div>
     <div class="people-list">
         <div class="header row padded">
-            <div class="column select-column">
-                <input type="checkbox" ng-model="$ctrl.selectAllValue" ng-change="$ctrl.selectAll()"/>
+            <div class="sticky-column col-bg-blue">
+                <div class="column select-column">
+                    <input type="checkbox" ng-model="$ctrl.selectAllValue" ng-change="$ctrl.selectAll()"/>
+                </div>
+                <!-- This is a duplicate of below. Angular is awful and doesn't let you make simple functions for stuff like this. -->
```
OzzieOrca: You have any suggestions for this? I suppose I could make a component...

adammeyer: Could you use an `ng-repeat`?

OzzieOrca: It is an ng-repeat. I just need to put the DOM for the first item somewhere else. Not sure what the best way to do that is.

adammeyer: Why are the two rows needed?  Sorry, I'm not able to figure that part out.

OzzieOrca: I want the mass select checkbox and the name column in a div so I can make them sticky. The rest can stay where they are. The first element of that columns array is the name column. Maybe there's a better way to structure it...

PR #32, opened 2018-08-14
better keyword errors

adammeyer:


app/assets/javascripts/angular/components/surveyOverviewKeywordTab/surveyOverviewKeyword.component.js
```diff
@@ -46,8 +63,13 @@ function surveyOverviewKeywordController(
...
                 },
-                () => {
-                    this.keywordError = true;
+                response => {
+                    const {
+                        data: { errors: [{ detail: details } = {}] = [] } = {},
+                    } = response;
+
+                    //display first error
+                    this.keywordError = details[_.keys(details)[0]][0];
```
OzzieOrca: Seems like some of this could be undefined. Also would calling `Object.values(details)[0]` be better?

adammeyer: Fixed

PR #34, opened 2018-08-15
remove unsafe JS links

adammeyer:
All have been tested in Safari


adammeyer:
I tested in Safari and Chrome, and functionality didn't change.  So I think we're good.

PR #35, opened 2018-08-16
csv valid types for windows

adammeyer:
Windows machines with Excel installed return a different type for the file.  This allows them to continue.


OzzieOrca:
Does that type include xlsx files?

adammeyer:
Yes, I believe it will, but I'm not sure of another way to fix that.

PR #40, opened 2018-08-22
Add ability to "Update SMS Keyword"

adammeyer:


adammeyer:
Fixes this bug: https://jira.cru.org/browse/MHP-1862

PR #42, opened 2018-08-24
MHP-1637: add predefined question modal

adammeyer:
MHP-1637

### DIFF
--- a/app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
+++ b/app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
@@ -99,6 +99,10 @@ function surveyOverviewQuestionsController($uibModal, surveyService) {
             });
     };
 
+    this.copyQuestion = question => {
+        this.addQuestion(_.omit(question, ['id']));
+    };
+
     this.deleteQuestion = questionId => {
         surveyService
             .deleteSurveyQuestion(this.survey.id, questionId)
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -67,9 +67,17 @@ function surveyOverviewQuestionsController(surveyService) {
...
+    };
 
-        surveyService.getPredefinedQuestions().then(questions => {
-            this.predefinedQuestions = questions;
+    this.addPredefinedQuestion = () => {
+        $uibModal.open({
+            component: 'predefinedQuestionsModal',
+            size: 'md',
+            resolve: {
+                addQuestion: _.constant(question => {
```
OzzieOrca: I'm not quite sure I understand all the function wrapping here but you could do `() => question => {` I think.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.html
```diff
@@ -63,7 +63,7 @@
                                ng-model="question.column_title" type="text">
                     </div>
                 </div>
-                <div class="flex justify-between">
+                <div class="flex justify-between" ng-if="question.kind === 'ChoiceField'">
```
OzzieOrca: This needs to be removed after we support auto assign and notify right?

adammeyer: Correct

PR #48, opened 2018-09-05
org management

adammeyer:


app/assets/javascripts/angular/components/orgManagement/orgManagement.scss
```diff
@@ -0,0 +1,566 @@
...
+
+    $wrapper-fixed-width: 1112px;
+
+    .wrapper {
+        width: $wrapper-fixed-width;
+        margin: 0 auto;
+        text-align: left;
+    }
+
+    .bg-pivot-gray-5 {
```
OzzieOrca: Can you remove this class and add it here https://github.com/CruGlobal/missionhub-web/blob/master/app/assets/stylesheets/variables/colors.scss#L43 ? That'll auto generate that class.

app/assets/javascripts/angular/components/orgManagement/orgManagementEdit.component.js
```diff
@@ -0,0 +1,41 @@
...
+function orgManagementEditController(organizationService) {
+    this.helpIcon = helpIcon;
+
+    this.$onInit = () => {
+        if (this.new) {
+            this.orgEdit = {
+                show_sub_orgs: true,
+            };
+        } else {
+            this.orgEdit = angular.copy(this.org);
```
OzzieOrca: `this.orgEdit = { ...this.org };` will do this too.

app/assets/javascripts/angular/components/orgManagement/orgManagementEdit.html
```diff
@@ -0,0 +1,55 @@
...
+    <h3 class="f3 fw6 pivot-dark-blue">
+        {{$ctrl.new ? 'common:organizations.new' : 'common:organizations.edit' | t}}
+    </h3>
+    <fieldset class="ba b--transparent ph0 mh0 w-100">
+        <div class="flex justify-between mb3">
+            <div class="w-48">
+                <label class="db f6 fw6 lh-copy normal mb1 pivot-grey-darker ttu tracked flex items-center"
+                       for="ministryName">
+                    {{'common:organizations.name' | t}}
+                    <a class="help" href="">
```
OzzieOrca: Is this supposed to go somewhere?

app/assets/javascripts/angular/components/orgManagement/orgManagementEdit.html
```diff
@@ -0,0 +1,55 @@
...
+                        <img alt="Help" ng-src="{{$ctrl.helpIcon}}" class="ml1 db" style="width:16px; height:16px;">
```
OzzieOrca: I feel like alt attributes should be translated but idk what the most efficient way to do that is...

app/assets/javascripts/angular/components/orgManagement/orgManagementSubOrg.component.js
```diff
@@ -0,0 +1,87 @@
...
+            organizationOverviewSuborgsService
+                .loadOrgSubOrgs(this.org)
+                .then(subOrgs => {
+                    this.subOrgs = subOrgs;
+                });
+        }
+    };
+
+    this.editSubOrg = ($event, orgId) => {
+        $event.stopPropagation();
```
OzzieOrca: Why is this needed?

app/assets/javascripts/angular/components/orgManagement/orgManagementSubOrg.component.js
```diff
@@ -0,0 +1,87 @@
...
+        this.newExpanded[orgId] = !this.newExpanded[orgId];
+    };
+
+    this.editComplete = (orgId, refresh) => {
+        this.editExpanded[orgId] = false;
+        this.newExpanded[orgId] = false;
+
+        if (refresh) {
+            this.orgExpanded[orgId] = false;
+            $timeout(() => {
```
OzzieOrca: This seems hacky...

app/assets/javascripts/angular/components/orgManagement/orgManagementSubOrg.component.js
```diff
@@ -0,0 +1,87 @@
...
+
+        confirmModalService
+            .create(
+                tFilter('ministries:organizations.delete.confirm', {
+                    org_name: org.name,
+                }),
+            )
+            .then(() => {
+                organizationService.deleteOrg(org).then(() => {
+                    this.subOrgs.splice(this.subOrgs.indexOf(org), 1);
```
OzzieOrca: A filter looking for all items that don't match that id might read better... Idk. This is fine. I guess a splice is probably more performant.

app/assets/javascripts/angular/components/orgManagement/orgManagementEdit.html
adammeyer: Alex is going to send me those.

app/assets/javascripts/angular/components/orgManagement/orgManagementSubOrg.component.js
adammeyer: So the parent h2 isn't clicked as well.

app/assets/javascripts/angular/components/orgManagement/orgManagementSubOrg.component.js
adammeyer: What do you recommend?  I thought of having another attribute on the sub-org component, that changed and have a watch inside the component to refresh the orgs on change.  That seems hacky too.  Thoughts?

OzzieOrca: Ya... idk. You could share some refresh function with the parent component like I did in my PR with that onInit thing... In React, you would fire a Redux action and update the global state store which would rerender the components as needed.

app/assets/javascripts/angular/components/orgManagement/orgManagement.scss
```diff
@@ -12,7 +12,7 @@ org-management {
     }
 
     .bg-pivot-gray-5 {
-        background-color: #f0f0f0;
+        background-color: $gray-5;
```
OzzieOrca: That's not what I meant. https://github.com/CruGlobal/missionhub-web/blob/master/app/assets/stylesheets/variables/colors.scss#L55 makes a class for every variable in the $pivot-colors variable. All you have to do is add a new key and it will make the class. That way all these classes stay centralized.

src/i18n/locales/en-US.js
```diff
@@ -562,7 +562,10 @@ export default {
             show_sub: 'Show Sub-Organizations',
             none_found: 'No organizations found',
             delete: {
-                confirm: 'Are you sure you want to delete "{{org_name}}"?',
+                confirm:
+                    'Are you sure you want to delete "{{org_name}}"?' +
+                    '<br><br>' +
```
OzzieOrca: This might be better as 2 different keys

app/assets/javascripts/angular/components/orgManagement/orgManagementSubOrg.component.js
adammeyer: I tried a few different refresh functions and wasn't able to come up with anything.  I'm not sure what I'm doing wrong, but it would always just refresh the parent, not the sub.

OzzieOrca: I guess this is fine. If you were doing a refresh function, you'd have to write the refresh function in the child component, call an output function on the child component $onInit, save a reference to it in the parent, and then pass it to this component to call. Angular makes this too complex...

PR #49, opened 2018-09-06
MHP-1932: remove required attribute

adammeyer:
https://jira.cru.org/browse/MHP-1932


adammeyer:
No, I don't believe so.  I'll remove that condition.

PR #56, opened 2018-09-20
MHP 1942 add new survey response model

jesusfreak3:
This is built off MHP-1674-survey-responses

### DIFF
--- a/app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
+++ b/app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
@@ -18,11 +18,11 @@ angular.module('missionhubApp').component('surveyResponseModal', {
     bindings: {
         dismiss: '&',
     },
-    controller: function($scope, $window) {
+    controller: function($scope, localStorageService) {
         var vm = this;
 
         $scope.closeModal = function() {
-            $window.localStorage.setItem('newSurveyResponseModal', true);
+            localStorageService.set('newSurveyResponseModal', true);
             vm.dismiss({ $value: 'cancel' });
         };
     },
@@ -33,11 +33,11 @@ function organizationOverviewSurveyResponsesController(
     surveyService,
     periodService,
     $uibModal,
-    $window,
+    localStorageService,
 ) {
     this.surveyStats = {};
     this.$onInit = () => {
-        if (!$window.localStorage.getItem('newSurveyResponseModal')) {
+        if (!localStorageService.get('newSurveyResponseModal')) {
             this.showSurveyModal();
         }
 
--- a/app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
+++ b/app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
@@ -18,7 +18,17 @@ function organizationOverviewSurveyResponsesController(
     periodService,
     $uibModal,
     localStorageService,
+    $log,
 ) {
+    let surveyResponseModalClosed = async modal => {
+        try {
+            await modal.closed;
+            localStorageService.set('newSurveyResponseModal', true);
+        } catch (error) {
+            $log.log(error);
+        }
+    };
+
     this.surveyStats = {};
     this.$onInit = () => {
         if (!localStorageService.get('newSurveyResponseModal')) {
@@ -36,11 +46,12 @@ function organizationOverviewSurveyResponsesController(
     };
 
     this.showSurveyModal = () => {
-        $uibModal.open({
+        const modal = $uibModal.open({
             component: 'surveyResponseModal',
             size: 'sm',
             windowClass: 'pivot_theme',
         });
+        surveyResponseModalClosed(modal);
     };
 
     this.getSurveyStats = () => {
--- a/app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
+++ b/app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
@@ -18,15 +18,10 @@ function organizationOverviewSurveyResponsesController(
     periodService,
     $uibModal,
     localStorageService,
-    $log,
 ) {
     let surveyResponseModalClosed = async modal => {
-        try {
-            await modal.closed;
-            localStorageService.set('newSurveyResponseModal', true);
-        } catch (error) {
-            $log.log(error);
-        }
+        await modal.closed;
+        localStorageService.set('newSurveyResponseModal', true);
     };
 
     this.surveyStats = {};
--- a/app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
+++ b/app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
@@ -1,4 +1,5 @@
 import template from './organizationOverviewSurveyResponses.html';
+import modalTemplate from './surveyResponsesModal.html';
 import './organizationOverviewSurveyResponses.scss';
 import _ from 'lodash';
 
@@ -12,13 +13,34 @@ angular
         controller: organizationOverviewSurveyResponsesController,
     });
 
+angular.module('missionhubApp').component('surveyResponseModal', {
+    template: modalTemplate,
+    bindings: {
+        dismiss: '&',
+    },
+    controller: function($scope, $window) {
+        var vm = this;
+
+        $scope.closeModal = function() {
+            $window.localStorage.setItem('newSurveyResponseModal', true);
+            vm.dismiss({ $value: 'cancel' });
+        };
+    },
+});
+
 function organizationOverviewSurveyResponsesController(
     $scope,
     surveyService,
     periodService,
+    $uibModal,
+    $window,
 ) {
     this.surveyStats = {};
     this.$onInit = () => {
+        if (!$window.localStorage.getItem('newSurveyResponseModal')) {
+            this.showSurveyModal();
+        }
+
         //get survey stats
         this.getSurveyStats();
 
@@ -26,6 +48,14 @@ function organizationOverviewSurveyResponsesController(
         periodService.subscribe($scope, () => {
             this.surveyStats = {};
             this.getSurveyStats();
+        });
+    };
+
+    this.showSurveyModal = () => {
+        $uibModal.open({
+            component: 'surveyResponseModal',
+            size: 'sm',
+            windowClass: 'pivot_theme',
         });
     };
 
--- a/app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
+++ b/app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
@@ -1,5 +1,4 @@
 import template from './organizationOverviewSurveyResponses.html';
-import modalTemplate from './surveyResponsesModal.html';
 import './organizationOverviewSurveyResponses.scss';
 import _ from 'lodash';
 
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/surveyResponsesModal.html
```diff
@@ -0,0 +1,8 @@
+<div class="modal-body form">
+    <p>You're experiencing the new version of Survey Responses here in MissionHub! Not quite ready for the update? Head over to your preferences and revert to Legacy. </p>
```
OzzieOrca: Since this is (hopefully) a short lived modal, I guess it doesn't need to be translated. But most other strings shown to the user in the app should be translated. You can use `t` from `import { t } from 'i18next';` in JS and the [`t` pipe](https://github.com/CruGlobal/missionhub-web/blob/master/app/assets/javascripts/angular/filters/translation.filter.js#L5) in templates.

You can add the translations in https://github.com/CruGlobal/missionhub-web/blob/master/src/i18n/locales/en-US.js and then provide the translation keys to the `t` function.

The top level keys in en-US.js are the namespace. Common is the default and doesn't need be specified. If you want another namespace the string becomes `t('myNamespace:translationKey.nestedTranslationKey')`. I've sorta been creating a namespace per component or group of related components. React lets you specify the namespace to use inside a component. The Angular libraries don't provide that as far as I can tell and you have to specify the namespace every single time (except for common which is the default).

There is currently tons of stuff in the common namespace as these translations were migrated from Rails. Ideally we should slowly move them out and only have VERY common strings like close, save, edit, etc in the common namespace.

Docs:
https://www.i18next.com
https://www.i18next.com/translation-function/essentials
https://www.i18next.com/translation-function/interpolation
https://www.i18next.com/principles/namespaces

app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/surveyResponsesModal.html
```diff
@@ -0,0 +1,8 @@
...
+</div>
+<div class="modal-footer">
+    <button class="btn btn-primary" type="button" ng-click="closeModal()">
```
OzzieOrca: Use `$ctrl.closeModal()` and put that function on `this` in the component. For components, the `this` of the component's controller gets mapped to `$scope.$ctrl`. The [`controllerAs`](https://docs.angularjs.org/api/ng/service/$compile#-controlleras-) value for components is `$ctrl`. See https://docs.angularjs.org/guide/component#comparison-between-directive-definition-and-component-definition

app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
```diff
@@ -18,11 +18,11 @@ angular.module('missionhubApp').component('surveyResponseModal', {
     bindings: {
         dismiss: '&',
     },
-    controller: function($scope, $window) {
+    controller: function($scope, localStorageService) {
         var vm = this;
 
         $scope.closeModal = function() {
-            $window.localStorage.setItem('newSurveyResponseModal', true);
+            localStorageService.set('newSurveyResponseModal', true);
```
OzzieOrca: I think I might prefer to keep the localStorage manipulation all in the same component. That way the localStorage code isn't separated and should be easier to find/follow. Idk. You could use the promise returned by `$uibModal` to do it.

app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/surveyResponsesModal.html
jesusfreak3: Thanks, yes that makes sense, lots of unknowns in this project right now since there was no real readme for standards etc, so will catch up on this bit by bit!  But  I have built my own translator services and filters for other projects for Angular like this, will correct this.

app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
```diff
@@ -12,13 +13,34 @@ angular
         controller: organizationOverviewSurveyResponsesController,
     });
 
+angular.module('missionhubApp').component('surveyResponseModal', {
```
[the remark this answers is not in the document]
jesusfreak3: Not a problem, i have worked in many other projects where the folder and file structure is so different (related components/services defined in same file, and able to be exported seperatly via ES6), and others with individual components etc per file.  I will keep to the 1 file per component as the standard for this project.  But yes, i do not agree with the structure of the site, putting all the folders under assets then js then angular (very stange... it is the application, so it should be in the root of app and not in a js folder??).  Then also the each component with its on folder... it has made it difficult to follow.  I think one day if possible, all the angular code should be moved to app folder, and then put in related folder structure, but thats another time/day/year.   I figured since this modal was directly related to this survey response it should be stored together, but will keep them in separate files, but same folder.

app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
```diff
@@ -12,13 +13,34 @@ angular
...
+    template: modalTemplate,
+    bindings: {
+        dismiss: '&',
+    },
+    controller: function($scope, $window) {
+        var vm = this;
```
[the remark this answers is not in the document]
jesusfreak3: I very really enjoy ES6, actually have converted many Angular 1 projects to VueJs, which really allows you to use the full power of ES6 and the new apis.  Its weird, I fire wrote the concept with no vm, and using $ctrl with shorthand function (arrow) but it would not work, then saw another example in your project using vm hack and scope and it worked.  I will try it again.

app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/surveyResponsesModal.html
jesusfreak3: Thanks, yes all my projects use the component functionality introduced in 1.3-1.5 (many also use the component routing) so I am very familiar with it, just hit issues as mentioned in the previous comment with the modal component.  But will adjust and try again.

app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
jesusfreak3: True, since the components will be in separate files as per your standard, it makes it harder to follow whats happening without opening the other component file.

jesusfreak3:
@OzzieOrca I am getting this "You should prefer the factory() method instead of service() " in the build, but why are we using this limitation? (Also why are we not pre linting this locally instead of finding out on the server build only?)  Service can be written the exact same way as a factory (like i am doing with return object ) however, there is one more benefit, if you really want to go more ES6, you can use ES6 Class with a service, but cannot with a factory. 

`class MyService {
  sayHello() {
    console.log('hello');
  }
}

app.service('MyService', MyService);`

 Let me know.

app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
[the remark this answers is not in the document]
jesusfreak3: Sorry not sure why, but yes it works as it should, arrow function with the closeModal.  Maybe i did someting stupid that day by putting an arrow function on the controller.. which wouldnt work as its a constructor...  anyways, take a look.

app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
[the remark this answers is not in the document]
OzzieOrca: Lol. Ya this code was pulled out the legacy Rails app. Someone had started writing Angular there... I couldn't stand it anymore and moved it to its own repo and webpack. That is the folder structure from the Rails asset pipeline still hanging around. I didn't want to make my big migration diff any bigger at the time. I guess I should have cleaned up the directories shortly after that but haven't. I was sorta hesitant cuz it would make git blame harder to use and I seemed to be constantly wondering why things were implemented in weird ways and trying to look up the original commits.

I like the one thing per file pattern. I think it can be hard to see where one component stops and another starts especially with Angular's config and especially when you get services with nested objects and lots of nested curly braces. It also makes files shorter. But I absolutely think things should be grouped into the same folders or put is subfolders so you can find related things. The original developer put everything in a flat folder structure which makes for a lot of things to scan through and makes it hard to find related components.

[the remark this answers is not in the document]
jesusfreak3: makes sense.  Lets hope one day soon the structure will be adjusted 👍 but will keep the one file per component logic, np.

app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
```diff
@@ -16,9 +16,25 @@ function organizationOverviewSurveyResponsesController(
     $scope,
     surveyService,
     periodService,
+    $uibModal,
+    localStorageService,
+    $log,
 ) {
+    let surveyResponseModalClosed = async modal => {
+        try {
```
OzzieOrca: Is this try needed? If there's an error it should* bubble up and get logged by rollbar.

I've also had issues with async/await and Angular 1.x. It seems to like to swallow promise rejections and never call the catch. Maybe that was only when using the api service... Were you having issues with the $log here? I think it might have something to do with not calling $digest after a promise completes like the $q implementation does.

app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/surveyResponsesModal.component.js
```diff
@@ -0,0 +1,13 @@
+import modalTemplate from './surveyResponsesModal.html';
+
+angular.module('missionhubApp').component('surveyResponseModal', {
+    template: modalTemplate,
+    bindings: {
+        dismiss: '&',
+    },
+    controller: function() {
+        this.closeModal = () => {
```
OzzieOrca: I think you could call $ctrl.dismiss directly from the template. Is the `{ $value: 'cancel' }` needed?

app/assets/javascripts/angular/services/localStorage.service.js
```diff
@@ -0,0 +1,19 @@
+angular.module('missionhubApp').service('localStorageService', () => {
+    const storage = {
+        set: (key, value) => {
+            localStorage.setItem(key, value);
```
OzzieOrca: Did we want to serialize anything we get here to JSON before storing?

OzzieOrca:
I changed the base branch of this PR initially so it was easier to review but I think it might be good to just merge it into my branch when it's ready so we can QA and deploy it all at once. I'm going need to update my survey responses branch but I'll wait to do this until this is in there.

app/assets/javascripts/angular/components/organizationOverviewSurveyResponses/organizationOverviewSurveyResponses.component.js
jesusfreak3: yah your right, i left it there, I was testing and debuging :S

app/assets/javascripts/angular/services/localStorage.service.js
jesusfreak3: For this case it was never the case, only using the fromJson to convert the string of 'true' into true since localstorage only stores in string.  I would assume if we wanted to store a json object in localstorage we would then stringify it before passing it to local storage instead of forcing the service to do it?  Not sure whats best

OzzieOrca: Seems like doing it in the service would be a general way to do it and would apply to all cases. I can't think of a downside. It should still serialize strings ok.

One fun note is that anything that comes from the JsonApiDataStore (which is what deserializes API responses) has the potential for circular references which JSON.stringify can't handle. Also JsonApiDataStore has a poorly implemented global store that I've been trying to avoid but was used extensively by previous developers.

jesusfreak3: Oh yes,... I have smacked my head a few times when trying to debug the data and receiving the circular references.  Yah I would be super happy to see that whole thing go, over complicated and weirdly implement...  About the Localservice store, sure I will just serialize the to json first, it should work for most cases as you  mentioned.

PR #59, opened 2018-09-27
MHP-1939: Add previously used questions to predefined questions modal

OzzieOrca:
Saving these currently throws an API error. I commented on https://jira.cru.org/browse/MHP-1939. Sheldon and I discussed this previously and I should have made an API ticket earlier.


app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -74,6 +74,7 @@ function surveyOverviewQuestionsController($uibModal, surveyService) {
             component: 'predefinedQuestionsModal',
             size: 'md',
             resolve: {
+                orgId: () => this.survey.organization_id,
```
jesusfreak3: Just a question, whats the standards for naming, I see full names for other resolve names, but orgId for this one.  I totally agree with short names for small scopes (IE in mapping, filters etc..) but was wondering about here

OzzieOrca: You talking about camelCase versus snake_case? Or shortening organizationId to orgId? Organization is used so much I guess I've been abbreviating it a bunch. The abbreviation should be pretty obvious. Idk what's best. You're right a lot of the components use the whole organizations name. I think I abbreviated it a bunch on the mobile app too cuz we were passing it around so much there. Idk. How do you feel about it? I sorta wish the API returned camelCase stuff...

OzzieOrca:
@jesusfreak3 Can you review my latest commit? I was made aware of a new API relation and a way to combine 2 requests. I based this off your branch cuz it needs the remove 3 predefined questions commit. We have some fun branch dependencies going on now :) Maybe we should merge all the finished survey stuff into a feature branch and then merge that later to deploy it all at once.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -260,7 +260,9 @@ function surveyOverviewQuestionsController(
             component: 'predefinedQuestionsModal',
             size: 'md',
             resolve: {
-                currentQuestions: () => this.surveyQuestions.map(q => q.id),
+                orgId: () => this.survey.organization_id,
+                currentQuestions: () =>
+                    this.surveyQuestions.map(question => question.id),
```
OzzieOrca: Looks like I changed this back in a merge conflict resolution... Looks like you shortened the name to fit it on one line. I don't care that much. I like meaningful names but this is trivial. I could undo that.

I realized a good practice might be that if we don't care enough about a variable to give it a meaningful name, maybe we should destructure it directly. I guess that could depend on context but it seems like it would work with simple things.
```suggestion
                    this.surveyQuestions.map(({ id }) => id),
```

jesusfreak3: yup i agree, the substructure is  a nice touch :)

jesusfreak3:
@OzzieOrca Looks good, makes sense reusable is both now predefined and previously used and we no longer really just need predefined anymore since they are in the surveys.  And yes I agree, we should merge all into 1 feature branch and coordinate with sheldon for the predefined and sort ones on his side.

PR #60, opened 2018-09-28
MHP-1973: Upgrade npm packages

OzzieOrca:


package.json
```diff
@@ -16,82 +16,97 @@
...
-        "prettier": "^1.13.5",
-        "pretty-quick": "^1.6.0",
-        "sass-loader": "^7.0.3",
-        "webpack": "^4.12.0",
-        "webpack-bundle-analyzer": "^2.13.1",
-        "webpack-cli": "^3.0.8",
-        "webpack-dev-server": "^3.1.4",
-        "webpack-manifest-plugin": "^2.0.3",
-        "webpack-subresource-integrity": "^1.1.0-rc.4"
+        "karma-webpack": "3.0.0",
```
OzzieOrca: `karma-webpack` didn't want to be updated. I spent long enough digging into it. The outputs object is empty and it tries to do a file operation with a file path of undefined. There are some PRs that supposedly fixed the errors I was seeing but they were more for the path being an array instead of undefined.

I guess it's fine stuck at `3.0.0` for now.

PR #66, opened 2018-10-17
Mhp 1594 implement autoassignment notification features

jesusfreak3:


app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -52,21 +61,162 @@ function surveyOverviewQuestionsController($uibModal, surveyService) {
...
+    const rebuildQuestion = question => {
+        let q = buildQuestion(question);
+        let index = this.surveyQuestions.indexOf(question);
+        this.surveyQuestions[index] = q;
+    };
+
+    const buildQuestion = question => {
+        let q = question;
+        let a = question.content ? question.content.split('\n') : [];
+        q.question_answers = a;
```
OzzieOrca: Could define the whole thing at once instead over a couple lines.
```suggestion
        const q = {
          ...question,
          question_answers: a,
        }
```

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -52,21 +61,162 @@ function surveyOverviewQuestionsController($uibModal, surveyService) {
...
+                a,
+                'AUTOASSIGN',
+            );
+            let autoNotifyRules = buildRules(
+                question.question_rules,
+                a,
+                'AUTONOTIFY',
+            );
+
+            q.question_rules = autoassignRules.concat(autoNotifyRules);
```
```suggestion
            q.question_rules = [...autoassignRules, ...autoNotifyRules];
```
works too.

app/assets/stylesheets/variables/generateColors.scss
```diff
@@ -11,6 +11,9 @@
     .b--pivot-#{$name} {
         border-color: $value;
     }
+    .bb--pivot-#{$name} {
+        border-bottom-color: $value !important;
```
OzzieOrca: ahh `!important` is unfortunate. That may be reality though...

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.html
```diff
@@ -16,13 +18,20 @@
...
+                ng-click="$event.stopPropagation()"
+                ng-change="$ctrl.saveQuestion(question)"
+                ng-model-options='{ debounce: 750 }'
+                ng-if="$ctrl.isExpanded[question.id]"
+                >
+            <a class="ml3" href=""
+                ng-if="$ctrl.isExpanded[question.id]"
+                ng-click="$ctrl.deleteQuestion(question.id)"
+                >
+                    <img ng-src="{{$ctrl.icons.trash}}" alt="{{ 'general:delete' | t }}" style="width:20px; height:24px;">
```
OzzieOrca: I found one of these cropped the other day when the width didn't match the height. Idk why a lot of icons in the app have a style attribute... whatever. If this looks fine then we're good.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.html
```diff
@@ -55,112 +63,117 @@
...
-                                <!--</div>-->
-                            <!--</div>-->
-                        <!--</div>-->
-                    <!--</div>-->
+                    <div class="w-48" ng-if="question.question_answers.length" ng-init="answerTab = 0">
+                        <div class="tabs flex justify-between">
+                            <div class="w-49 relative">
+                                <a class="ttu b f6 bl br bt pv2 ph2 link pivot-gray-darker br2 br&#45;&#45;top db b&#45;&#45;gray-dark"
+                                   href="" ng-click="answerTab = 0" ng-class="{'bb bb--pivot-gray-5': answerTab === 0}">
+                                    Auto Assign
```
OzzieOrca: i18n
```suggestion
                                    {{ 'someNamespace:autoAssign' | t }}
```

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.html
```diff
@@ -55,112 +63,117 @@
...
-                                        <!--</li>-->
-                                    <!--</ul>-->
-                                <!--</div>-->
-                            <!--</div>-->
-                        <!--</div>-->
-                    <!--</div>-->
+                    <div class="w-48" ng-if="question.question_answers.length" ng-init="answerTab = 0">
+                        <div class="tabs flex justify-between">
+                            <div class="w-49 relative">
+                                <a class="ttu b f6 bl br bt pv2 ph2 link pivot-gray-darker br2 br&#45;&#45;top db b&#45;&#45;gray-dark"
```
OzzieOrca: What is `&#45;` doing here? Seems like it snuck in from a copy/paste or something.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.html
```diff
@@ -55,112 +63,117 @@
...
+                                    Auto Assign
+                                </a>
+                                <a class="help fr p2" href="">
+                                    <img alt="Help" ng-src="{{$ctrl.icons.help}}" style="width:16px; height:16px;">
+                                </a>
+                            </div>
+                            <div class="w-49 relative">
+                                <a class="ttu b f6  bl br bt pv2 ph2 link pivot-gray-darker br2 br&#45;&#45;top db b&#45;&#45;gray-dark"
+                                   href="" ng-click="answerTab = 1" ng-class="{'bb bb--pivot-gray-5' : answerTab === 1}">
+                                    Auto Notify
```
OzzieOrca: i18n

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.html
```diff
@@ -55,112 +63,117 @@
...
+                                </a>
+                                <a class="help fr" href="">
+                                    <img alt="Help" ng-src="{{$ctrl.icons.help}}" style="width:16px; height:16px;">
```
OzzieOrca: i18n the alt Help attribute. I think there's an example in the repo somewhere. I assume `alt="{{ 'help' | t}}"` will work but maybe not...

app/assets/javascripts/angular/services/survey.service.js
```diff
@@ -131,14 +134,56 @@ function surveyService(
...
                     type: 'question',
                     attributes: attributes,
                 },
             };
 
+            if (rules) {
+                let ruleRelationship = [];
+                let ruleIncludes = [];
+
+                rules.forEach(r => {
```
OzzieOrca: Could you do a reduce? Something like
```js
const { ruleRelationship, ruleIncludes } = rules.reduce((acc, rule) => ({
  ruleRelationship: {
    ...acc.ruleRelationship,
    ...condition ? [newStuff] : []
  },
  ruleIncludes: {
    ...acc.ruleIncludes,
    ...condition ? [newStuff] : []
  }
}), { ruleRelationship: [], ruleIncludes: [] } );
```
Maybe that's too many things happening at once but I kinda like it.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -52,21 +61,162 @@ function surveyOverviewQuestionsController($uibModal, surveyService) {
...
+                ar.organization_ids = r.organization_ids;
+                ar.people_ids = r.people_ids;
+                ar.trigger_keywords = r.trigger_keywords;
+
+                if (!r.people_ids) {
+                    return;
+                }
+
+                let ids = r.people_ids.split(',');
+                ar.assignTo = this.people.filter(p => ids.indexOf(p.id) >= 0);
```
OzzieOrca: Is this supposed to be `assign_to`? The casing is different. Maybe you're using this locally for something.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -15,25 +15,34 @@ angular.module('missionhubApp').component('surveyOverviewQuestions', {
     template: template,
 });
 
-function surveyOverviewQuestionsController($uibModal, surveyService) {
+function surveyOverviewQuestionsController(
```
OzzieOrca: This got to be a pretty big component. Anything that belongs in a service or helper JS file?

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -52,21 +61,162 @@ function surveyOverviewQuestionsController($uibModal, surveyService) {
...
+            {
+                errorMessage: 'error.messages.surveys.loadQuestions',
+            },
+        );
+
+        return r.data;
+    };
+
+    const rebuildQuestions = questions => {
+        this.surveyQuestions = questions.map(q => buildQuestion(q));
```
OzzieOrca: Does 
```suggestion
        this.surveyQuestions = questions.map(buildQuestion);
```
work?

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -52,21 +61,162 @@ function surveyOverviewQuestionsController($uibModal, surveyService) {
...
+
+                let ids = r.people_ids.split(',');
+                ar.assignTo = this.people.filter(p => ids.indexOf(p.id) >= 0);
+            }
+        });
+
+        return ar;
     };
 
     this.$onInit = () => {
```
OzzieOrca: I like keeping $onInit near the top so you can find it. Idk how it got all the way down here.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -52,21 +61,162 @@ function surveyOverviewQuestionsController($uibModal, surveyService) {
...
-        surveyService.getSurveyQuestions(this.survey.id).then(questions => {
-            this.surveyQuestions = questions;
+        loadSurveyData();
+    };
+
+    this.addPersonToRule = async (question, rule) => {
+        let index = question.question_rules.indexOf(rule);
```
OzzieOrca: If you modify `rule` directly, does it not update the parent array? Seems like it should since it's a reference.

I do like striving for immutable patterns but I know that's not always easy, especially in Angular.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -52,21 +61,162 @@ function surveyOverviewQuestionsController($uibModal, surveyService) {
...
+        }
+
+        let ids = [
+            ...new Set(question.question_rules[index].assignTo.map(r => r.id)),
+        ];
+        let currentIds = question.question_rules[index].people_ids
+            ? question.question_rules[index].people_ids.split(',')
+            : [];
+
+        if (_.isEqual(currentIds.sort(), ids.sort())) {
```
OzzieOrca: I wish the JS Set API had more methods available.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -52,21 +61,162 @@ function surveyOverviewQuestionsController($uibModal, surveyService) {
...
+            return;
+        }
+
+        question.question_rules[index].people_ids = ids.join(',');
+        question.question_rules[index].assignTo.forEach(a => {
+            let exists = this.people.find(p => p.id === a.id);
+            if (!exists || exists === undefined) {
+                this.people.push(a);
```
OzzieOrca: What is this doing? Seems like you're duplicating people in `this.people`. Find a person in people that matches an assignment and duplicate that person onto the end of the list. I don't really if `a` is a whole person object or what.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -52,21 +61,162 @@ function surveyOverviewQuestionsController($uibModal, surveyService) {
...
+                    }
+
+                    return a2 + r.people_ids + ',';
+                }, '')
+            );
+        }, '');
+
+        let peopleIds = [...new Set(list.split(','))];
+
+        let r = await httpProxy.get(
```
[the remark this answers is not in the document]
jesusfreak3: Yah would still need another line, but its more elegant your way and clearer i think.

jesusfreak3:
"Most of my comments are stylistic anyways. "  Thats fine, I am always happy to see improvements stylistic.

"I like trying to keep things immutable and a byproduct of that is trying to use a map or reduce instead of forEach (which usually results in extra mutations) if possible. "

So was a little confused about this, as I used 2 foreach through the entire code, and the rest are maps/filters/reduce.  But i think the one example you saw right away was "buildRule" which looking at it, I see a filter and map could have done the same thing, Ill rework that.

Going through the rest.

app/assets/stylesheets/variables/generateColors.scss
jesusfreak3: Yah i hate using important, but the way the css is setup with the pivitol theme etc, the other classes they have setup are way to specific instead of general for reuse, so they will override any general classes... its a pain...   Alex told me there are other tickets for cleaning up and fixing the css, so that is where all this stuff will need to be resolved.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
jesusfreak3: Yah its strictly local to hook into the assign lookup components. It is not sent to the server. But to keep standards the same I can switch it to underscore from camel case. I think originally i had it as a local variable only which did not work, then moved it into the object. Its one of those annoying things, variables functions camelcase (which i like) and then entity/objects from apis underscores.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
jesusfreak3: Yah i agree.. what about private methods, do you want them below or above.  Init got pushed down as i moved the private const methods to the top (normally keep the private methods separated from the pubic) as oninit is public.  Let me know how you want this.  SO oninit is the first public method.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
jesusfreak3: Yah, but the one handy thing it does, is it doesnt insert duplicates which is why i am using it here for the ids, to avoid duplicate ids being added.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
OzzieOrca: Gotcha. Just checking cuz it looked like `assign_to` was a thing too. Ya the API using snake case annoys me.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
jesusfreak3: So here I need to update the people list with any new person that is not pulled originally on init to create the label.  So when the user adds a person via the assign component, i then check to see if this person is already in our person list, if not then push it in so the labels work (first and last name).  Yah "a" is the assigned person object returned from the component, its a partial object, pretty much first, last and email etc.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.html
jesusfreak3: Yah, i really dislike style attributes, these icons came from the original html, and saw it done like this in other places so i left it.  Its also the same for the help icons.. just saw it, and if you remove the style attributes... it changes nothing...  So stupid.. should have caught that and removed them!  Ill do that now.

OzzieOrca:
Sorry if that was confusing. PRs seem to point out everything to be improved and usually don't do a  good job of celebrating what looks good. (I guess I could do better with that.)

I was excited to see you using map and reduce a lot of other places 😄 I think those forEach loops stood out when I was going through it partly cuz of the complexity surrounding them and cuz I was trying to think how a reduce might fit there. That doesn't mean you can't use a forEach, I just like trying to see how functional programming patterns fit. I'm never sure if I should comment about that or not. It does take some buy in.

Maybe I put too many comments here... Thanks for going through them :)

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
OzzieOrca: Idk that I do a very good job of separating private and public. I kinda like grouping functionality together. Maybe I should think about that more. Structure it however you want as long as there's some sort of thought put into it.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.html
jesusfreak3: ahah.. yah need to put that in the back of my mind about the translations. thx

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
OzzieOrca: Ya that's awesome. I only discovered that recently. Just wishing it did more now :)

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.html
OzzieOrca: Good to know. I should have tried that a few places. Idk what the original intention was.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.html
jesusfreak3: good catch, yah general.help will push soon

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
OzzieOrca: Oh does not exist. I didn't read that exists condition right. Isn't `!exists || exists === undefined` the same as `!exists`?

jesusfreak3: Yah, i thought you missed the not exist! :P 

from my memory undefined == false returns false, so need to check for undefined.  while I was watching output, i saw it returning undefined sometimes.  So added the extra condition?

OzzieOrca: `!undefined === true` is true. I don't think the seperate check is needed unless I'm missing something.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.html
jesusfreak3: Nope, i thought that too originally, so i changed it back to - but everytime you commit it looks like your precommit feature is converting these to html codes??

jesusfreak3:
I rebuilt the build rule to avoid the mutation and use map filter and reduce etc.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.html
OzzieOrca: That's very strange. It should be ignoring html files https://github.com/CruGlobal/missionhub-web/blob/master/.prettierignore#L7.  I think there's a git flag to skip hooks. You could try that. I can look into it more if you want. Just not sure where to go with it. Guess I could try reproducing it.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -148,34 +148,38 @@ function surveyOverviewQuestionsController(
...
+                    id: r.id,
+                    label_ids: r.label_ids,
+                    organization_ids: r.organization_ids,
+                    people_ids: r.people_ids,
+                    rule_code: type,
+                    trigger_keywords: answer,
+                    assign_to: assignTo,
+                };
+            })
+            .reduce((a, c) => (c ? c : false), false);
```
OzzieOrca: Is this essentially a get last? Seems like a find/findLast might be better to find the rule you want and then just manipulate that one rule. Maybe I'm missing something. Just realized the original code did this too although the `assignTo` seems like it could be assigned in a previous loop and then not cleared out when you find a new one.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -52,21 +61,162 @@ function surveyOverviewQuestionsController($uibModal, surveyService) {
...
+                    if (!r.people_ids) {
+                        return a2;
+                    }
+
+                    return a2 + r.people_ids + ',';
+                }, '')
+            );
+        }, '');
+
+        let peopleIds = [...new Set(list.split(','))];
```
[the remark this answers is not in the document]
jesusfreak3: About char variable names, yah I never liked that before, i always named variables with the full long names, until i met some other developers, and they convinced me about scope and naming.  Function signatures should always use full naming, but why make the code so much longer in maps, and filters, etc  when the scope is so small, and you know map outputs the value from the array which should be named already properly.  SO i just make sure i use chars that make sense, ie, if pulling questions in a map etc i use q, if its rules, then r, etc..

IE

`const rule = questionRules.filter(r => r.trigger_keywords === answer && r.rule_code === type)`

That could be written like this 

`const rule = questionRules.filter(rule => rule.trigger_keywords === answer && rule.rule_code === type)`

but thats extra typing, and longer to read that isnt necessary as we know the map has to return the item (rule) in the array named questionRules.

what do u think

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
jesusfreak3: yup it does, nice catch.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
OzzieOrca: I don't think this reduce does anything other than return the last item of the array right? You're always returning something in the map so the false part of `c ? c : false` never gets hit.

jesusfreak3: No, the rule may not exist,  so then we need a "empty" rule placeholder for the UI to work.  So find will return false if nothing is found, and then the map will fail as you cannot map a non array.  Pretty much the UI expects us to show the find a person for each keyword, but the API does not have a rule setup for each keyword (question.context separated by ",")  So then I need to push the empty placeholder rule with null id to the UI so the user can use it when they want to add the rule.  Its a limitation to the API data due to legacy data, and the required UI look.  Does that make sense?

jesusfreak3: > I don't think this reduce does anything other than return the last item of the array right? You're always returning something in the map so the false part of `c ? c : false` never gets hit.

your right can just write 
`.reduce((a, c) => (c), false);`

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
jesusfreak3: yup, actually just refactored a bit more, dont even need q, and no longer mutation :)  Yah the ... is beautful, arrays and objects...

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
OzzieOrca: I'm still confused. Seems like you just need one loop to find the rule you want if it exists. Is there something else you're doing here?

Something like this might work. (Idk why I keep writing code. I started and then wanted to have something to share I guess. I wasn't trying to write code for you... You can take it as an example.)

```js
const {
  id,
  label_ids,
  organization_ids,
  people_ids,
  assign_to
} =
  questionRules.find(
    r => r.trigger_keywords === answer && r.rule_code === type
  ) || {};

const ids = people_ids ? people_ids.split(',') : [];

return {
  id,
  label_ids,
  organization_ids,
  people_ids,
  rule_code: type,
  trigger_keywords: answer,
  assign_to: id ? this.people.filter( p => ids.indexOf(p.id) >= 0) : null
};
```
If you need null instead of undefined you could do:
```js
const {
  id = null,
  label_ids = null,
  ...
} = ...
```

Hopefully that's right... It may not be...

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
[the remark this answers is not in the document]
OzzieOrca: Guess I just always like having context so you can look through it quickly. Especially when writing code in complex loop I forget what the variables are representing and don't like to have to look above.  What you have is fine. Once you have more than one or 2 variables, I think it gets hard to read.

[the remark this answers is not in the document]
jesusfreak3: Cool, ill keep that in mind, about the 1 or 2 variables. thx

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
jesusfreak3: ok i see, yah i can do somthing like that, I still need to pass an empty rule if the find returns false, but that will work.

OzzieOrca: That's what the ` || {};` was for.

OzzieOrca: I forgot to delete `assign_to` up top in my code...

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -338,15 +338,12 @@ function surveyOverviewQuestionsController(
...
-            const ruleIndex = question.question_rules.findIndex(
-                r => r.trigger_keywords === changedAnswer,
-            );
-
             const rules = question.question_rules.map(r => {
                 if (r.trigger_keywords === changedAnswer) {
-                    let rule = r;
-                    rule.trigger_keywords = newAnswer;
-                    return rule;
+                    return (rule = {
```
OzzieOrca: What's the `rule = ` for? Seems like it just goes out of scope immediately.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -118,28 +263,107 @@ function surveyOverviewQuestionsController($uibModal, surveyService) {
...
+        this.saveQuestionContent(question, answers);
+    };
+
+    this.updateQuestionAnswer = async (question, answerIndex) => {
+        const oldAnswers = question.content.split('\n');
+        const newAnswer = question.question_answers[answerIndex];
+
+        if (oldAnswers.indexOf(newAnswer) === -1) {
+            const changedAnswer = oldAnswers.find(
+                a => question.question_answers.indexOf(a) === -1,
```
OzzieOrca: What does this do? Seems like it just finds the last oldAnswer that is also a new answer. How do you know that's the one that changed?

jesusfreak3: oldAnswers is the array of the previous answers before change, so then i find which of the new answers dont match in the old answers, which means that is the changed answer.

OzzieOrca: Oh right. I'm not reading your conditions very well today... I was thinking that was the opposite. Maybe a comment there would be good. This part is a little confusing.

PR #67, opened 2018-10-17
Mhp 1588 implement reordering survey questions

jesusfreak3:


app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -219,6 +219,26 @@ function surveyOverviewQuestionsController(
...
+    };
+
+    this.updatePosition = questions => {
+        let i = 1;
+        questions.filter(q => !q.predefined).forEach(q => {
+            q.position = i;
+
+            let { id, position } = q;
+
+            surveyService.updateSurveyQuestion(this.survey.id, {
```
OzzieOrca: Do we make an API request for every question each time any question is reordered. Seems like a lot of requests.

Did anything ever happen with [this comment](https://jira.cru.org/browse/MHP-1588?focusedCommentId=70004&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-70004)?

> Scotty also mentioned to me that because the design uses a drag-and-drop UI for ordering we will need to find a way to reshuffle the position of each element when one is moved. This could be a bulk update endpoint, or we could try to auto update those positions magically in the API.

If you could just send the updated position for one question and the API could figure out how to shuffle them around, that seems better. If this is what you have settled on with the API devs then it's fine.

Seems like that might have been implemented https://jira.cru.org/browse/MHP-1850?focusedCommentId=72268&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-72268. You might only have to save the one that changed and the API will reshuffle them. Idk what that means for keeping the local positions in sync with the API though...

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.html
```diff
@@ -11,9 +14,9 @@
                 'pa3': !$ctrl.isExpanded[question.id],
                 'ph3 pv2 items-center': $ctrl.isExpanded[question.id]
         }">
-            <!--<a class="sort mr3">-->
-                <!--<img ng-src="{{$ctrl.icons.sort}}" alt="Sort" style="width:22px; height:20px">-->
-            <!--</a>-->
+            <a class="question-sort mr3" ng-if="!question.predefined">
+                <img ng-src="{{$ctrl.icons.sort}}" alt="Sort" style="width:22px; height:20px">
```
OzzieOrca: i18n

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
jesusfreak3: Yah, after i pushed i followed up with scott about that feature.  I built it so the drag and drop just passed up the new position for each question which did the same thing, but going to adjust to use the APIs ability.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -219,6 +219,26 @@ function surveyOverviewQuestionsController(
...
+                id,
+                position,
+            });
+
+            i++;
```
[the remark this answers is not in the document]
jesusfreak3: duuhh.. i think i was sleeping when i added the manual indexing... 🥇

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
jesusfreak3: ok updated logic here, it will only save questions where the position changed instead of all of them.  However, if you take question number 1 and move it to the end, it will say all the question positions were modified.   The component we used just updated the list of objects into the new sort, it doesnt tell you hey i moved x to y, so then from the new sort you can see what indexes moved.

PR #68, opened 2018-10-18
Fix mass entry modal button text color while saving

jesusfreak3:


jesusfreak3:
lol yah, i saw it and was like hmm.. is it a trick???  But yah same i had the assumption they wanted to replace the text with like saving or a animated icon.

PR #71, opened 2018-10-25
Mhp 1628 create menu for unauthenticated users

jesusfreak3:
You will need to checkout the same branch in missionhub legacy which i added you as a PR reviewer too.


jesusfreak3:
You can run this application outside of rails, but you still need rails running on port 3000 as we are still using css from that.  I have a ticket for that to clean it up and then we wont need it.  But when you run it manually you need to update the jwt-token in index.ejs with the one running locally in rails (you can see it in the auth headers, or in the html rendered of the preload component.

app/assets/javascripts/angular/components/nav/nav.html
```diff
@@ -53,9 +53,23 @@
...
+
+    <ul ng-if="!$ctrl.loggedInPerson.person" class="sf-menu mh-top" id="primary_nav">
+        <li class='ml-auto'>
+            <a href="http://help.missionhub.com" target='_blank'>{{ 'nav.support' | t }}</a>
+        </li>
+        <li>
+            <a ui-sref="app.login">{{ 'nav.login' | t }}</a>
+        </li>
+        <li>
+            <a href="http://localhost:3000/request_access">{{ 'nav.requestAccess' | t }}</a>
```
OzzieOrca: This needs to change depending on environment. `href="/request_access"` might work everywhere except running outside of Rails. Looks like you added a `railsUrl` to the env config too.

app/assets/stylesheets/pivot_theme.scss
```diff
@@ -8,15 +7,80 @@
...
+    background-color: $gray-5 !important;
 }
 
 body.white-background {
     background: white !important;
 }
 
+.main-container {
+    margin: 0 auto;
+    width: 976px;
```
OzzieOrca: Whoever initially did this was evil lol. Just chucks responsive design out the window.

app/assets/stylesheets/pivot_theme.scss
```diff
@@ -8,15 +7,80 @@
...
+}
+
+.btn-primary {
```
OzzieOrca: Maybe some of these could go in a seperate CSS file? Not that this main CSS file is small by any means but I think keeping as much code out of it as possible would be good.

app/assets/stylesheets/pivot_theme.scss
```diff
@@ -8,15 +7,80 @@
...
+    -webkit-appearance: none;
+    border-radius: none;
+    -moz-border-radius: none;
+    -webkit-border-radius: none;
+    text-align: center;
+}
+
+.btn-key {
+    &:before {
+        background: url(http://localhost:3000/assets/welcome/logo-key-1x.png)
```
OzzieOrca: Needs to change per environment. Seems like moving this image over to the new repo and using a relative path would be an easy fix.

app/assets/stylesheets/pivot_theme.scss
```diff
@@ -8,15 +7,80 @@
...
+        position: relative;
+        top: 2px;
+        left: 0;
+        float: left;
+    }
+}
+
+.btn-facebook {
+    &:before {
+        background: url(http://localhost:3000/assets/welcome/facebook-1x.png)
```
OzzieOrca: Needs to change per environment. Seems like moving this image over to the new repo and using a relative path would be an easy fix.

app/assets/stylesheets/pivot_theme.scss
```diff
@@ -8,15 +7,80 @@
...
+    }
+}
+
+.topmenu {
+    margin-bottom: 40px;
+    height: 64px;
+    background: $pivot-darker-blue;
+    display: -webkit-box;
+    display: -webkit-flex;
+    display: -ms-flexbox;
```
OzzieOrca: Do we need all these prefixes anymore?

index.ejs
```diff
@@ -5,14 +5,25 @@
...
 
-    <title>Survey</title><!-- TODO: make title dynamic -->
-
+    <title>MissionHub</title>
+    <link rel="stylesheet" type="text/css" href="http://localhost:3000/assets/dashboard.css?body=1">
+    <link rel="stylesheet" type="text/css" href="http://localhost:3000/assets/global.css?body=1">
+    <link rel="stylesheet" type="text/css" href="http://localhost:3000/assets/superfish.css?body=1">
+    <link rel="stylesheet" type="text/css" href="http://localhost:3000/assets/superfish-navbar.css?body=1">
+    <link rel="stylesheet" type="text/css" href="http://localhost:3000/assets/mh_permissions_labels_style.css?body=1">
+    <link rel="stylesheet" type="text/css" href="http://localhost:3000/assets/mh_profile_style.css?body=1">
```
OzzieOrca: Sounds like you're moving these in a different ticket. Just don't want them going to prod with those urls.

jesusfreak3: This wont affect prod, anything in the index js file will only load via angualr app, if you load via rails, this index file is ignored.  SO this is for testing while moving everything over.

app/assets/javascripts/angular/components/nav/nav.html
jesusfreak3: Good catch, yah i replaced all the others with the new railsUrl from env, will fix this one.

app/assets/stylesheets/pivot_theme.scss
jesusfreak3: yes, indeed.  the next tickets for fix app css is going to deal with all this

app/assets/stylesheets/pivot_theme.scss
jesusfreak3: the next tickets for fix app css is going to deal with all this

app/assets/stylesheets/pivot_theme.scss
jesusfreak3: yup, cant believe i forgot this.... :S

app/assets/stylesheets/pivot_theme.scss
jesusfreak3: Ok i fixed this, and removed a few other places as well.

app/assets/stylesheets/pivot_theme.scss
```diff
@@ -17,9 +17,29 @@ body.white-background {
...
-    width: 976px;
+    margin-right: auto;
+    margin-left: auto;
+    padding-left: 15px;
+    padding-right: 15px;
+
+    min-height: calc(100vh - 140px /* root navigation */ - 43px /* footer */);
+
+    @media screen and (min-width: 768px) {
+        width: 750px;
```
OzzieOrca: Does this actually work well everywhere? I'd assumed nested things might be hardcoded... I can play with it some.

jesusfreak3: YAh it works for the most part, I fixed a lower down section too, there is still one level of responsivness that needs to be adjusted, once you get below 672px, there is some off screening that needs to be resolved on the dashboard and the tables start having issues around 539px.  But wouldnt take much to make it work all the way.  Might be able to takle it in the fix css stuff.

PR #72, opened 2018-11-01
Mhp 1617 Remove dependency on rails CSS

jesusfreak3:
This is not 100% ready, it works, everything have been removed from rails and into its own, but not in a separate scss file yet.  But go a head and start the review since i do not work friday.  Then I can review it next week and also split of the scss


app/assets/javascripts/angular/components/nav-search/nav-search.component.js
```diff
@@ -7,7 +7,7 @@ angular.module('missionhubApp').component('navSearch', {
 });
 
 function navSearchController(peopleSearchService) {
-    var vm = this;
-
-    vm.searchPeople = peopleSearchService.search;
+    this.$onInit = () => {
+        this.searchPeople = peopleSearchService.search;
```
OzzieOrca: Does this need to be in an `$onInit`? Seems like the injected service should be available...

app/assets/javascripts/angular/components/nav-search/nav-search.html
```diff
@@ -1,9 +1,9 @@
-<div id="menu_search">
+<div id="">
```
OzzieOrca: You can just toss the id

app/assets/stylesheets/pivot_theme.scss
```diff
@@ -8,13 +7,274 @@
 @import './components/navigation';
 @import './components/row';
 @import './components/sheet';
+@import '~tachyons';
+
+//This is the new section here
+
+app {
```
OzzieOrca: I haven't reviewed any of this. I'll let you clean it up first. I assume a lot of it was copy/pasted from the old repo.

src/components/login.html
```diff
@@ -0,0 +1,23 @@
+<div class="m-auto tc">
+    <h1 class="ttu">{{ 'login.login' | t }}</h1>
+    <p>
+        <a href
+            class="btn-key btn-primary f5"
+            title="{{ 'login.login_with_key' | t }}"
+            target="_self"
```
OzzieOrca: Is this needed? Maybe the router config is still weird since most of it is under `/d/`...

src/components/navHeader.html
```diff
@@ -0,0 +1,76 @@
+<sunset-legacy-env-bar></sunset-legacy-env-bar>
+<div class="main-menu bg-pivot-darker-blue">
+
+    <a href="/" class='logo fl' ng-if="$ctrl.state.branded"></a>
+
+    <ul class="font-titillium fl ml3 list" ng-if="$ctrl.state.hasMissionhubAccess && $ctrl.loggedInPerson.person" id="primary_nav">
```
OzzieOrca: There are a lot of floats here. Flex might be better but idk that you need to rewrite it. I'm not sure if you wrote the HTML or pulled it from somewhere.

app/assets/javascripts/angular/components/nav-search/nav-search.html
jesusfreak3: yah gota stop doing that, i remove it while i am testing to make sure i got all the css.. :S

src/components/login.html
jesusfreak3: Sorry is which part still needed?

OzzieOrca: Just wondering what `target="_self"` does. I don't usually see it used.

jesusfreak3: when you run the app locally outside of rails, its no longer under /d/.  If your asking about the _self, you use target self when you use relative urls, but dont want it to go through the router but as if its an external url with out opening a new window.  Is that what you were asking?

OzzieOrca: Interesting. But you've got an absolute url. I guess that's because it's the same domain but different port, maybe? Makes more sense after seeing these:
https://stackoverflow.com/a/28802951/665224
https://docs.angularjs.org/guide/$location#html-link-rewriting
If it doesn't work without it, keep it :) I was mainly just curious since I hadn't really seen `_self` used before.

jesusfreak3: np ;) Yah i have had to use it in other AngularJs SPs when they had mixed legacy stuff.  But you are write, I changed it in the end to use the railurl so its an absolute url now instead of relative like original.

app/assets/stylesheets/pivot_theme.scss
jesusfreak3: yah a mix, i rewrote a bunch of it, and copied some of it.

jesusfreak3: ok revamped a lot, moved into components and also into separate scss files and removed not needed code.

src/components/navigation/navHeader.scss
```diff
@@ -0,0 +1,123 @@
...
+    left: 0;
+    right: 0;
+    z-index: 100;
+
+    .main-menu {
+        height: 64px;
+        padding: 15px 15px;
+
+        .logo {
+            background: url(/app/assets/images/logo-missionhub-initials.png)
```
OzzieOrca: I've seen webpack do weird things with absolute image urls (like not even add the image to the list of output files), especially in a production build. If it works on Netlify that's a good enough test. If it's got issues a relative url might be better.

jesusfreak3: Its rails, rails is rewritting the url that I am attempting to access in the app.  Will see what I can do about that.

jesusfreak3: So the off center nav is because of the missing brand logo...  You can see here locally how it shows and is aligned, so will fix that logo issue.

image

jesusfreak3: Looking into the ng cloak, doesnt happen locally through rails, but only on staging, so checking that.

jesusfreak3: Ok the images are fixed now.

jesusfreak3: Ok no more flash is occuring, and the ministry link is fixed too.

jesusfreak3: also its more responsive now, it will at least adjust down to tablet size etc.  Mobile is still an issue untill we spend some time into the grid sections in contact etc..

OzzieOrca:
Code looks good. Profile picture is still square. Menu flashes to unauthenticated version before showing the authenticated version when you refresh the page.

jesusfreak3:
I pushed the avatar rounded issue in 1614 since its based off this branch.  There is no more flashing of non auth links on refresh that was fixed a while back.  so I think we are good here.

PR #74, opened 2018-11-15
MHP-2094: Add public survey preview

OzzieOrca:


app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -418,8 +418,14 @@ angular
             })
             .state({
                 name: 'publicSurvey',
-                url: '/s/:surveyId',
+                url: '/s/:surveyId?preview',
                 component: 'publicSurvey',
+                resolve: {
+                    survey: ($state, $transition$, routesService) =>
+                        routesService.getSurvey($transition$.params().surveyId),
```
jesusfreak3: Hey so what is $transition$ and why use that instead of stateParams?

OzzieOrca: I have no idea. I could try it again. I can never remember UI-Router's API and looking it up is a nightmare. The 2 versions of it don't help. I copied this from somewhere in our app.

I have a faint memory that `stateParams` wasn't working but idk why. I don't see stateParams used anywhere in missionhubApp.routes.js

jesusfreak3: Sorry yah, its cool Transition.

PR #75, opened 2018-11-15
Mhp 1614 authentication

jesusfreak3:
Added The key authentication, and logout.  This will no longer run within rails, so run locally with yarn only.


app/assets/javascripts/angular/services/localStorage.service.js
```diff
@@ -13,6 +36,15 @@ angular.module('missionhubApp').service('localStorageService', () => {
         clear: () => {
             localStorage.clear();
         },
+        allowSessionTransfer: () => {
```
OzzieOrca: Can you add some docs for this? This is kinda a weird feature and some future developer is going to stumble upon this and be confused.

app/assets/javascripts/angular/services/localStorage.service.js
```diff
@@ -1,4 +1,27 @@
...
+                angular.toJson(sessionStorage),
+            );
+            localStorage.removeItem('sessionStorage');
+        } else if (
+            newEvent.key === 'sessionStorage' &&
+            !sessionStorage.length
+        ) {
+            const data = angular.fromJson(newEvent.newValue);
+
+            for (var key in data) {
```
OzzieOrca: `let`?

Isn't `data` a [StorageEvent](https://developer.mozilla.org/en-US/docs/Web/API/StorageEvent)? I think it contains data for a single key that was changed. I'm trying to figure out why you are looping over the event object's keys. I think the keys you'll get out of your loop are are key, newValue, oldValue, storageArea, and url. I feel like I'm missing something.

index.ejs
```diff
@@ -5,14 +5,20 @@
...
+      integrity="sha256-xNjb53/rY+WmG+4L6tTl9m6PpqknWZvRt0rO1SRnJzw="
+      crossorigin="anonymous"></script>
 </head>
 
-<body ng-cloak ng-app="missionhubApp">
-
-    <ui-view></ui-view>
-
+<body class='bg-pivot-gray' ng-cloak ng-app="missionhubApp">
+    <app></app>
```
OzzieOrca: Yay 🎉

package.json
```diff
@@ -67,6 +67,7 @@
         "@babel/plugin-syntax-dynamic-import": "^7.0.0",
         "@babel/plugin-syntax-import-meta": "^7.0.0",
         "@babel/plugin-transform-runtime": "^7.0.0",
+        "@babel/plugin-transform-template-literals": "^7.0.0",
```
OzzieOrca: Was this not working before?

app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -0,0 +1,158 @@
...
+        const me = await loggedInPerson.loadOnce();
+
+        i18next.changeLanguage(me.user.language);
+        updateRollbarPerson(me);
+        $rootScope.legacyNavigation = me.user.beta_mode === false;
+    };
+
+    const setHttpHeaders = token => {
+        if (!token) return;
+        $http.defaults.headers.common.Authorization = 'Bearer ' + token;
```
OzzieOrca: Do we need to clear this in `clearToken`?

OzzieOrca:
Looks like all the builds failed cuz of an eslint error.

I want to try and get this running on Netlify at some point. Currently I'm redirecting the index.html file away to make the current Netlify links work... Also feel like there are going to be some CORS issues.

app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -249,12 +243,70 @@ angular
...
+                                        resolve: {},
+                                        windowClass: 'pivot_theme',
+                                        size: 'sm',
+                                    });
+                                }
+
+                                deferred.resolve(user);
+                            })
+                            .catch(e => {
+                                if (e.status === 401) {
```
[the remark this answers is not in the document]
jesusfreak3: Ill check with Sheldon to see if the API returns anything else for this type of response.

app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -249,12 +243,70 @@ angular
...
+
+                        return deferred.promise;
                     },
                 },
                 template: '<ui-view></ui-view>',
             })
+            .state({
+                name: 'login',
+                url: '/sign-in',
+                component: 'login',
```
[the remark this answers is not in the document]
jesusfreak3: Lol yes,  domain driven is better option, if users/client are used to sign in more, then lets use that.   It depends on where your from lol.   In the ticket shannon or whoever created it uses the key and facebook login, but then says implement sign out, but in the auth create sign in landing page... lol  I got confused for which one to use... lol

app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -249,12 +243,70 @@ angular
...
+            })
+            .state({
+                name: 'auth',
+                url: '/auth?acces_token',
+                component: 'login',
+                resolve: {
+                    accessToken: $location => {
+                        const token = $location
```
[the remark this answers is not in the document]
jesusfreak3: YAh, i started with the router ui API for trying to get the hash via the route params,. and after many attempts and lack of documentation, i gave up and wrote that.   I tried playing with regex part, but it never allowed me to use it with hash urls, just path based and query string based.   I also noticed MPDX project ended up doing a similar solution to mine for parsing, same with the examples that Nate sent me.  Wasn't sure how much time they wanted me to waste on this parser.  Ill put it in some helper for parsingHashUrls or something.

app/assets/javascripts/angular/services/localStorage.service.js
jesusfreak3: IS there a standard way for documentation here for this, or your talking about inline comments?

package.json
jesusfreak3: Now im not sure, ill re test, i thought it wasnt so I added it.  Ill remove it and retest.

jesusfreak3:
Ok I upgraded yarn and checked out the yarn.lock file and re ran yarn, let me know if it resolved the issue.

app/assets/javascripts/angular/services/localStorage.service.js
jesusfreak3: So yes you are correct, we are pulling the values of the key we called "sessionStorage" which we created when the event to open a new tab.  So this sessionStorage value includes all the key/values for what we had in the sessionStorage at that moment.  In this case, its the below.

jwtToken: ""xxxxxxxxxxxxxxxxxxxxxxxxxx""
state: "{"hasMissionhubAccess":true,"currentOrganization":"10046"}"

app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -0,0 +1,158 @@
...
+            false
+        );
+    };
+
+    const setupUserSettings = async () => {
+        const me = await loggedInPerson.loadOnce();
+
+        i18next.changeLanguage(me.user.language);
+        updateRollbarPerson(me);
+        $rootScope.legacyNavigation = me.user.beta_mode === false;
```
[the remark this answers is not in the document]
jesusfreak3: Yah that was the last reference to legacy that I found, all others were removed, so I can quickly remove this too, i meant to ask you about it!?

app/assets/javascripts/angular/missionhubApp.routes.js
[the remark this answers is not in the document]
jesusfreak3: sheldon says that 401 should be good.

app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -0,0 +1,158 @@
...
+        state.currentOrganization = currentState.currentOrganization;
+        state.loggedIn = true;
+
+        $rootScope.$broadcast('state:changed', state);
+    };
+
+    const setState = organization => {
+        const newState = {
+            hasMissionhubAccess: true,
+            currentOrganization: organization ? organization.id : 0,
```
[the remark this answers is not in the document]
jesusfreak3: Yah I was basing this on the logic from the legacy rail app.

app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -0,0 +1,158 @@
...
+    const theKeyloginUrl = `${envService.read(
+        'theKeyUrl',
+    )}/login?response_type=token&scope=fullticket&state=randomData&client_id=${envService.read(
+        'theKeyClientId',
+    )}&redirect_uri=${redirectUrl}`;
+
+    const getJwtToken = () => {
+        return (
+            sessionStorageService.get('jwtToken') ||
+            localStorageService.get('jwtToken') ||
```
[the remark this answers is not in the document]
jesusfreak3: right your saying if somoene managed to inject the jwtToken then we would be pulling from it when not on prod.  Sure it may not be needed, but safer then sorry.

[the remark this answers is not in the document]
jesusfreak3: done

app/assets/javascripts/angular/services/authentication.service.js
jesusfreak3: Yah better  safe then sorry.  Someone could be logged out and if someone attempted via consol etc to make a request they would still be valid.

jesusfreak3: done

app/assets/javascripts/angular/services/authentication.service.js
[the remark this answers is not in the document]
OzzieOrca: Ya you can remove it. I don't think we have any plans to keep any sort of a hybrid with links back to Rails.

app/assets/javascripts/angular/services/localStorage.service.js
OzzieOrca: I think just add some inline comments near there...

app/assets/javascripts/angular/services/localStorage.service.js
OzzieOrca: Sorry I missed the `.newValue` and the fact that you had packed them all into the `sessionStorage` key. You're good.

OzzieOrca:
Your new commits look good :) I marked a bunch of the comments as resolved. There are still some  others but most of them are code cleanup, not functionality. Still wondering about `randomData` in the key url though.

app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -0,0 +1,158 @@
...
+            code: ticket,
+        });
+
+    const authorizeAccess = async accessToken => {
+        const response = await requestTicket(accessToken);
+        const { data } = await requestV4Token(response.data.ticket);
+
+        setAuthorizationAndState(data.token, data.recent_organization);
+
+        $state.go('app.people');
```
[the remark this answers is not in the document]
jesusfreak3: Added the handling in the Facebook ticket for this.

app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -0,0 +1,158 @@
...
+            const token = getJwtToken();
+
+            if (!token) return;
+
+            setHttpHeaders(token);
+            loadState();
+        },
+        theKeyloginUrl: theKeyloginUrl,
+        isTokenValid: () => {
+            return getJwtToken();
```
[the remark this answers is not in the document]
jesusfreak3: sure,

app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -0,0 +1,158 @@
...
+    localStorageService,
+    loggedInPerson,
+) {
+    const service = `${envService.read('apiUrl')}/auth/thekey`;
+    const redirectUrl = encodeURIComponent(
+        envService.read('siteUrl') + '/auth',
+    );
+    const theKeyloginUrl = `${envService.read(
+        'theKeyUrl',
+    )}/login?response_type=token&scope=fullticket&state=randomData&client_id=${envService.read(
```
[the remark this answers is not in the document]
jesusfreak3: apparently not lol... its removed.

app/assets/javascripts/angular/services/authentication.service.js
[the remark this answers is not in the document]
jesusfreak3: removed.

app/assets/javascripts/angular/missionhubApp.routes.js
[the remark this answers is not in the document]
jesusfreak3: Refactored into simple service

app/assets/javascripts/angular/services/localStorage.service.js
jesusfreak3: cool done.

app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -0,0 +1,158 @@
...
+    const loadState = () => {
+        const currentState =
+            sessionStorageService.get('state') ||
+            localStorageService.get('state');
+
+        if (!currentState) return;
+
+        state.hasMissionhubAccess = currentState.hasMissionhubAccess;
+        state.currentOrganization = currentState.currentOrganization;
+        state.loggedIn = true;
```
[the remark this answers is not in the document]
jesusfreak3: Now the state should only be changed via the authentication service compared to before where it was changed in other locations.  But yah if we attempt to return a new state object it fails since its injecting it.  What we could do is update the state so that it has a parameter for the state values and then we can use new value objects via exposed methods.  But not sure if its worth all the hassle?  Let me know.

package.json
jesusfreak3: Ok removed it, and recompiled still works.  So I must have been sleeping when i did this!

app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -249,12 +243,70 @@ angular
...
+                                deferred.resolve(user);
+                            })
+                            .catch(e => {
+                                if (e.status === 401) {
+                                    authenticationService.removeAccess();
+                                    $state.go('login');
+                                }
+                            });
+
+                        return deferred.promise;
```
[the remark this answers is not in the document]
jesusfreak3: Ok, refactored, and moved authentication check into transtion of router ui. so that every time you switch routes it rechecks if you have access.

app/assets/javascripts/angular/missionhubApp.run.js
```diff
@@ -29,4 +29,25 @@ angular
...
+                )
+            )
+                return true;
+
+            if (!authenticationService.isTokenValid()) {
+                authenticationService.removeAccess();
+                return transition.router.stateService.target('signIn');
+            }
+
+            return loggedInPerson.loadOnce().catch(e => {
```
OzzieOrca: So is this going to run once on every page transition? I think the loadOnce name means it's cached. Might be good to do
```js
if (e.status === 401) {
  authenticationService.removeAccess();
  return transition.router.stateService.target('signIn');
}
```
in the API service somewhere so if we start getting errors deep in the app, we make them sign in again.

app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -25,6 +25,8 @@ function authenticationService(
         'theKeyClientId',
     )}&redirect_uri=${redirectUrl}`;
 
+    const nonAuthenticatedRoutes = ['signIn', 'auth'];
```
OzzieOrca: I think I've seen people put these in the router config before. Like
```js
.state({
  name: 'signIn',	
  url: '/sign-in',
  component: 'signIn',
  allowUnauthenticated: true,
})
```
My guess is you could look that up with `transition.to().allowUnauthenticated`. Maybe you can come up with a better name.

I think since we are going to have more public routes (public survey, join community, and maybe something with invite links) having this in the state config is more maintainable.

app/assets/javascripts/angular/services/authentication.service.js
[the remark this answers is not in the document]
OzzieOrca: Idk if it's worth the hassle. I think immutable things are easier to follow though.

app/assets/javascripts/angular/missionhubApp.run.js
jesusfreak3: Right, yah that would cache it, changed to load instead of loadOnce.  Yah there is no real other way in SP to make sure your still authenticated unless you check in between each transition/route.  This covers most cases.  But your right there is still a chance that there is data pulled/pushed with out a route change.  However, there could be a chance a user does an action they are not authorized for, but are still logged in, so catching a 401 at any point could be dangerous?

jesusfreak3: yah sheldon confirmed, there are places where 401 is returned even if the user is authenticated, so we cant inject it deep into the API

app/assets/javascripts/angular/services/authentication.service.js
jesusfreak3: Yah, makes more senses, done.  Added isPublic, so by default a route is not, so no need to add to all routes, just the ones you want to be able to access by the public.

app/assets/javascripts/angular/missionhubApp.run.js
OzzieOrca: I guess ideally you'd get a 401 if you are not authenticated or your authentication has expired. A 403 error would mean your authentication is good but you are not authorized to access that endpoint. I guess if the API isn't implemented like that, we can't do that.

Looks like mobile is having to rely on error messages...
https://github.com/CruGlobal/missionhub-react-native/blob/master/src/actions/api.js#L99-L117
https://github.com/CruGlobal/missionhub-react-native/blob/master/src/constants.js#L64-L66

Let's merge this without this and we can make a ticket for it. I'm a little worried that users will leave the tab open and start having requests fail without them forcing them to sign in again.

We probably don't need to use refresh tokens and can just force another sign in but I guess that could still be a discussion.

PR #77, opened 2018-11-20
Mhp 1615 facebook login

jesusfreak3:
This is based off 1614 branch.


app/assets/javascripts/angular/missionhubApp.run.js
```diff
@@ -24,9 +26,37 @@ angular
...
+                return;
+            }
+
+            let js = d.createElement('script');
+            js.id = id;
+            js.async = true;
+            js.src = '//connect.facebook.net/en_US/sdk.js';
+
+            ref.parentNode.insertBefore(js, ref);
+        })(document);
```
OzzieOrca: Can you stick all this and the `$window.fbAsyncInit` in the `facebookService`? You should just be able to call an initialization function here that will run all of this.

app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -89,12 +89,24 @@ function authenticationService(
...
     const authorizeAccess = async accessToken => {
         const response = await requestTicket(accessToken);
         const { data } = await requestV4Token(response.data.ticket);
 
         setAuthorizationAndState(data.token, data.recent_organization);
+        $state.go('app.people');
+    };
+
+    const authorizeFacebookAccess = async accessToken => {
+        const { data } = await requestFacebookV4Token(accessToken);
```
OzzieOrca: Again, if there is some API error here we may not want to `$state.go('app.people');` unless that redirects gracefully back to login if auth is missing.

app/assets/javascripts/angular/missionhubApp.run.js
jesusfreak3: ok take a look

package.json
```diff
@@ -2,7 +2,7 @@
     "name": "missionhub",
     "version": "1.0.0",
     "scripts": {
-        "start": "webpack-dev-server",
+        "start": "webpack-dev-server --https ",
```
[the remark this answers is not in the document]
jesusfreak3: done

app/assets/javascripts/angular/services/authentication.service.js
jesusfreak3: Ok so the error service would not necessarily catch everything, so i added try catch and then called the error service.

PR #78, opened 2018-11-22
MHP-2068: Add edit answer sheet modal, and add logic to call for editing from h…

jesusfreak3:
…istory

### DIFF
--- a/app/assets/javascripts/angular/components/answerSheet/editAnswerSheetModal.component.js
+++ b/app/assets/javascripts/angular/components/answerSheet/editAnswerSheetModal.component.js
@@ -1,0 +1,47 @@
+import template from './editAnswerSheetModal.html';
+
+import './editAnswerSheetModal.scss';
+
+angular.module('missionhubApp').component('editAnswerSheetModal', {
+    bindings: {
+        resolve: '<',
+        dismiss: '&',
+        close: '&',
+    },
+    template: template,
+    controller: editAnswerSheetModalController,
+});
+
+function editAnswerSheetModalController(httpProxy, modelsService) {
+    this.save = async answerId => {
+        this.resolve.answerSheet.answers.forEach(async a => {
+            let { data } = await httpProxy.put(
+                modelsService
+                    .getModelMetadata('answer_sheet')
+                    .url.single(this.resolve.answerSheet.id),
+                {
+                    data: {
+                        type: 'answer_sheet',
+                        attributes: {},
+                    },
+                    included: [
+                        {
+                            type: 'answer',
+                            attributes: {
+                                question_id: a.question.id,
+                                value: a.value,
+                            },
+                        },
+                    ],
+                },
+                {
+                    errorMessage: 'surveyTab:errors.updateSurvey',
+                },
+            );
+        });
+
+        this.dismiss();
+    };
+
+    this.onInit;
+}
--- a/app/assets/javascripts/angular/components/answerSheet/editAnswerSheetModal.component.js
+++ b/app/assets/javascripts/angular/components/answerSheet/editAnswerSheetModal.component.js
@@ -31,7 +31,7 @@ function editAnswerSheetModalController(httpProxy, modelsService) {
             .url.single(this.resolve.answerSheet.id);
 
         const { data } = await httpProxy.put(
-            `${url}?include=answers`,
+            url,
             {
                 data: {
                     type: 'answer_sheet',
@@ -40,6 +40,7 @@ function editAnswerSheetModalController(httpProxy, modelsService) {
                 included: includedAnswers,
             },
             {
+                params: { include: 'answers' },
                 errorMessage: 'surveyTab:errors.updateSurvey',
             },
         );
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/angular/components/answerSheet/editAnswerSheetModal.component.js
```diff
@@ -0,0 +1,47 @@
...
+                        },
+                    ],
+                },
+                {
+                    errorMessage: 'surveyTab:errors.updateSurvey',
+                },
+            );
+        });
+
+        this.dismiss();
```
OzzieOrca: I think `close` is generally used for a successful modal action. `dismiss` is for when a user backed out of taking whatever action the modal presented. It probably doesn't  matter in this case as we aren't using the result but the convention is good.

app/assets/javascripts/angular/components/answerSheet/editAnswerSheetModal.component.js
```diff
@@ -0,0 +1,47 @@
...
+            let { data } = await httpProxy.put(
+                modelsService
+                    .getModelMetadata('answer_sheet')
+                    .url.single(this.resolve.answerSheet.id),
+                {
+                    data: {
+                        type: 'answer_sheet',
+                        attributes: {},
+                    },
+                    included: [
```
OzzieOrca: I think you can pack a bunch of answers in this included array so you don't have to use a forEach to generate many requests. I think you can make this a single request but I could be wrong.

OzzieOrca:
The checkboxes also don't seem to load the labels correctly.
screen shot 2018-11-26 at 1 47 12 pm

app/assets/javascripts/angular/components/answerSheet/editAnswerSheetModal.component.js
jesusfreak3: Your right, didnt think much about it, but its true, usually the dimiss with do a reject on a given promoise and the close will do the resolve on the promise.

app/assets/javascripts/angular/components/answerSheet/editAnswerSheetModal.component.js
jesusfreak3: Your right, still getting used to this API.  Updated to save all at once.  I did find it weird that I could not send all, but missed the little list part in the API..  I had to add extra logic, as somewhere in the httpProxy, it rebinds the results, but the last result is missing related data, so it removes it.  So I simply just took the answers that we changed put them in a const, and rebound them after the await.  This resolve the issue with the last answer missing in the UI after the save.

OzzieOrca:
What is this https://github.com/CruGlobal/missionhub-web/pull/78/commits/2d4f9731759b6a577bb606591a63293c5f951c92#diff-7d3a395baa8934682f8ec41dd54297b4R50 ?

I'm a little confused by what's happening with JsonApiDataStore. You just restoring what it used to before the modification?

jesusfreak3:
Hey, yah so it maybe the API causing the issue, but i am restoring the data from before sending to the API, so it includes all the users changes.  The issue is the Data store rebinds the data from the returned API request and somewhere it loose part of the data to the last question answer, so then the UI removes the last question after the save.  The data was sent correctly and the API does save the changes, its just on the rebind, something funny happens, so i just rebind manually with the changes I sent to the API which will only occur on a success request.

OzzieOrca:
JsonApiDataStore should die...

You can probably add the includes you want to the API request so it returns them. Does sticking the query param `include=answers` work?

Idk why JsonApiDataStore is clearing out its store though. You'd think it would keep past data...

If this isn't any easy fix, just keep what you have. Maybe add a comment cuz it's weird...

jesusfreak3:
your right, added the include fixed the issue... Very weird behavior that the rebinding only caused problems for the last item in the list past to the API.

app/assets/javascripts/angular/components/answerSheet/editAnswerSheetModal.component.js
```diff
@@ -0,0 +1,51 @@
...
+            },
+            {
+                errorMessage: 'surveyTab:errors.updateSurvey',
+            },
+        );
+
+        this.close();
+    };
+
+    this.onInit;
```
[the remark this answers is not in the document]
jesusfreak3: yah i have been trying to figure out how i put it there.. its not even the correct format for $onInit.... must have been the new allowed marijuana regulations in Canada. :P I must have started writing on Init then forgot about it and your prettier added the ; as i dont even write semi colons, your prettier does that for me. 🥇

app/assets/javascripts/angular/components/answerSheet/editAnswerSheetModal.component.js
```diff
@@ -26,10 +26,12 @@ function editAnswerSheetModalController(httpProxy, modelsService) {
...
 
+        const url = modelsService
+            .getModelMetadata('answer_sheet')
+            .url.single(this.resolve.answerSheet.id);
+
         const { data } = await httpProxy.put(
-            modelsService
-                .getModelMetadata('answer_sheet')
-                .url.single(this.resolve.answerSheet.id),
+            `${url}?include=answers`,
```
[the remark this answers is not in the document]
jesusfreak3: looks like i redid that part anyways at one point

app/assets/javascripts/angular/components/answerSheet/editAnswerSheetModal.component.js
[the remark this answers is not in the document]
OzzieOrca: lol

PR #79, opened 2018-12-04
Mhp 2096 inactive surveys public

jesusfreak3:


jesusfreak3:
obviously just check since commit 5cc6cdb since branched of existing feature branch

app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -421,12 +421,29 @@ angular
                 url: '/s/:surveyId?preview',
                 component: 'publicSurvey',
                 resolve: {
-                    survey: ($state, $transition$, routesService) =>
-                        routesService.getSurvey($transition$.params().surveyId),
+                    survey: ($state, $transition$, routesService) => {
+                        try {
+                            routesService.getSurvey(
```
OzzieOrca: Doesn't this need a return?

app/assets/javascripts/angular/components/surveyOverview/surveyOverview.html
```diff
@@ -23,7 +23,7 @@
         </div>
     </div>
     </div>
-    <a ng-href="{{$ctrl.surveyLinkPrefix}}{{$ctrl.survey.id}}?preview"
+    <a ng-href="/d/previewSurvey/{{$ctrl.survey.id}}"
```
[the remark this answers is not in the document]
jesusfreak3: Yah its weird, I tried to get ui sref to work, but no avail.  Ill try somthing else, as I do agree, hardcoded /d/ is a problem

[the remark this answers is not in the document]
jesusfreak3: found the issue, originally was not going though app.  so was not going through ui view so injeciton etc were not handled so ui sref would not work.

app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -421,25 +421,30 @@ angular
...
-                    survey: ($state, $transition$, routesService) => {
-                        try {
-                            routesService.getSurvey(
-                                $transition$.params().surveyId,
-                            );
-                        } catch (e) {
-                            //bypass built in error handler that stops the routes
-                        }
+                    survey: ($state, $transition$, routesService, $q) => {
+                        const deferred = $q.defer();
```
OzzieOrca: I don't think you need defer. Something like this should work:
```js
return routesService
  .getSurvey($transition$.params().surveyId)
  .catch(e => null);
```
I think you have to rethrow the error in the catch if you want it to propagate otherwise the promise will just resolve.

app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -421,25 +421,30 @@ angular
...
                 },
             })
             .state({
-                name: 'previewSurvey',
+                name: 'app.previewSurvey',
                 url: '/previewSurvey/:surveyId',
                 component: 'publicSurvey',
                 resolve: {
-                    survey: ($state, $transition$, routesService) =>
+                    survey: ($state, $transition$, routesService, $q) =>
```
OzzieOrca: The `$q` looks unused.

app/assets/javascripts/angular/missionhubApp.routes.js
jesusfreak3: well isnt that stupid... i started that way at the start, but it didnt work, thought the httpproxy behind was still swolloing the error... But just retested, and it works.... 

image

OzzieOrca:
Looks like the input CSS is broken in the preview. Idk if that's worth fixing. Might just be better to wait till your other CSS lands to fix it.
screen shot 2018-12-04 at 6 53 17 pm

Also there's this hidden preview button. Idk why it's in there but it is.
screen shot 2018-12-04 at 6 56 25 pm

jesusfreak3:
yah exactly, its the global.css that is killing the padding on the inputs, so it will be resolved as soon as the new version is pushed.  I didnt think it mattered to much for the preview.

jesusfreak3:
Good catch on the hidden preview button, was wondering if there were more, didnt try those three dots menu... Pushed the fix

PR #82, opened 2018-12-11
Mhp 2033 add last survey to survey responses

jesusfreak3:


app/assets/javascripts/angular/components/peopleScreen/peopleScreen.html
```diff
@@ -23,7 +23,14 @@
                      ng-if="$first"
                      ng-class="[column.cssClass, { 'pointer-disabled': !column.sortable}]"
                      ng-click="column.sortable && $ctrl.setSortColumn(column)">
-                    <span>{{column.label | t}}</span>
+                    <span
+                        ng-if="
+                            column.name !== 'lastSurvey' ||
+                            ($ctrl.surveyId && column.name === 'lastSurvey')
```
OzzieOrca: What's this logic for? You hiding the label in some cases?

app/assets/javascripts/angular/components/peopleScreen/peopleScreen.html
```diff
@@ -110,7 +124,8 @@
                     selected="$ctrl.multiSelection[person.id]"
                     person="person" organization-id="$ctrl.org.id"
                     questions="$ctrl.questions"
-                    odd="$odd">
+                    odd="$odd",
+                    show-last-survey="$ctrl.surveyId !== false && $ctrl.surveyId !== undefined">
```
OzzieOrca: Does this work and cover more falsy values?
```suggestion
                    show-last-survey="$ctrl.surveyId">
```

app/assets/javascripts/angular/components/ministryViewPerson/ministryViewPerson.component.js
```diff
@@ -52,6 +53,14 @@ function ministryViewPersonController(
             },
         );
 
+        if (this.showLastSurvey) {
+            $scope.$watchCollection('$ctrl.person.answer_sheets', () => {
+                if (!this.person.answer_sheets) vm.lastSurvey = null;
+
+                vm.lastSurvey = personService.getLastSurvey(this.person);
```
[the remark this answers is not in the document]
jesusfreak3: hmm.. to be honest i didnt think much about it, i just implemented the same logic as is done for the other bindings being passed and watched.  You are totally right, the onChanges within a component is a better option.  The reason why its watching is that the value could change being passed to the component, so if we unwateched it after and it changed we would not get it.  But yes, onChanges is better.  Do you want this changed, if so, then it would make sense to change it for all the watchers as they are all on binding properties.

app/assets/javascripts/angular/components/peopleScreen/peopleScreen.html
jesusfreak3: I just realised, I could just not add the column to the column list in the controller if no survey id is present, then this logic would not need to exist in the view, what do you think.

app/assets/javascripts/angular/components/peopleScreen/peopleScreen.html
jesusfreak3: well... that just makes me look stupid ... thanks for that.

app/assets/javascripts/angular/components/peopleScreen/peopleScreen.html
OzzieOrca: I think that sounds better.

app/assets/javascripts/angular/components/ministryViewPerson/ministryViewPerson.component.js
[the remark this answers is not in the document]
OzzieOrca: Gotcha. Idk how much it matters... If you want to change them all, go for it, but this is probably fine.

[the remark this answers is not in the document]
jesusfreak3: Yah i cleaned it up, i think onChanges is more efficient then the multiple watchers.

app/assets/javascripts/angular/components/peopleScreen/peopleScreen.html
jesusfreak3: ok done. take a look

jesusfreak3:
ok should be good now.

PR #83, opened 2018-12-12
Mhp 2064 migrate analytics

jesusfreak3:


jesusfreak3:
@OzzieOrca ok so this is done, but I am not passing the SSOUID currently as I am waiting for API to add.  SO instead i am just placing a text placeholder.  Once the API returns it I will simply pass it into the placeholder.

app/assets/javascripts/angular/services/analyticsService.js
```diff
@@ -0,0 +1,93 @@
...
+            setupGoogle(ssoUid);
+            setupAdobe(ssoUid);
+        },
+        track: transition => {
+            const newState = transition.$to();
+            const currentUrl = newState.path.reduce((acc, p) => {
+                if (p.self.url === '^') return acc;
+
+                const url = Object.entries(transition.params()).reduce(
+                    (acc, param) => {
```
OzzieOrca: I like naming the things that come off of Object.entries. `param[0]` just seems like a magic number. I'm not sure what these represent so I named them param1 and param2 but a better name would be great :)
```suggestion
                    (acc, [param1, param2]) => {
```

app/assets/javascripts/angular/services/analyticsService.js
```diff
@@ -0,0 +1,93 @@
...
+                        return acc.replace(`:${param[0]}`, param[1]);
+                    },
+                    p.self.url,
+                );
+
+                return acc + url;
```
OzzieOrca: Not necessary, but I've grown to like template literals more than normal string concat.
```suggestion
                return `${acc}${url}`;
```

app/assets/javascripts/angular/services/analyticsService.js
```diff
@@ -0,0 +1,93 @@
...
+            }, '');
+
+            const fields = {
+                page: currentUrl,
+                location: envService.read('siteUrl') + currentUrl,
```
OzzieOrca: Could use template literal.

src/app.component.js
```diff
@@ -17,6 +17,8 @@ function appController(
     state,
     sessionStorageService,
     $timeout,
+    $analytics,
+    envService,
```
OzzieOrca: You using these?

app/assets/javascripts/angular/services/analyticsService.js
OzzieOrca: Where is `url` defined?

jesusfreak3:
Created new ticket, so i guess if its good we can merge this into the feature branch.

OzzieOrca:
Yep. Can you at least answer my `url` question? Looks like it may be undefined.

jesusfreak3:
Sorry, i dont know how but i didnt see any of your other question comments!  Ill check those.

app/assets/javascripts/angular/services/analyticsService.js
jesusfreak3: Yah, that makes more sense.  I keep forgetting all the great ways to use the destructive assignment...

app/assets/javascripts/angular/services/analyticsService.js
jesusfreak3: ahah, i had it like that as i had more "text" in it, then put it to simple concatenation after as the literals is alot more code for the same thing.  I have still been battling which i like more for simple two variable concatenation.

app/assets/javascripts/angular/services/analyticsService.js
jesusfreak3: same with this..

src/app.component.js
jesusfreak3: I was but forgot to clean it up... bah..

app/assets/javascripts/angular/services/analyticsService.js
jesusfreak3: I am confused, url is defined directtly above it

`const url = Object.entries(transition.params()).reduce(
                    (acc, param) => {
                        return acc.replace(`:${param[0]}`, param[1]);
                    },
                    p.self.url,
                );

                return acc + url;`

OzzieOrca: I'm dumb. Guess my comment was in the way... I wasn't reading the nested reduce right.

OzzieOrca: K, ya doesn't matter. I'm fine with either.

jesusfreak3: ahaha finally ":P

OzzieOrca:
Looks like you broke Travis. Everything else looks good

PR #84, opened 2018-12-18
Remove Netlify redirect to make Angular app run on its own

OzzieOrca:
- Cleanup env config
- Add new The Key client id that supports Netlify domains


app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -17,10 +18,7 @@ function authenticationService(
     errorService,
 ) {
     const service = `${envService.read('apiUrl')}/auth/thekey`;
-    const redirectUrl = encodeURIComponent(
-        envService.read('siteUrl') +
-            (envService.is('production') ? '/d/auth' : '/auth'),
-    );
+    const redirectUrl = encodeURIComponent(`https://${$location.host()}/auth`);
```
jesusfreak3: I assume that when this goes live we will no longer be using the /d/ url as other wise this change will fail on production.

jesusfreak3:
@OzzieOrca looks good, just that one question.

PR #87, opened 2018-12-20
update to pass GR id, FB id, and SSOuid for adobe.  Pass SSOUId also …

jesusfreak3:
Ok here it is, now have the ids passed to adobe etc..

### DIFF
--- a/app/assets/javascripts/angular/missionhubApp.run.js
+++ b/app/assets/javascripts/angular/missionhubApp.run.js
@@ -32,13 +32,8 @@ angular
 
         facebookService.loadSDK()(document);
 
-        loggedInPerson.loadOnce().then(user => {
-            analyticsService.init(
-                user.thekey_uid,
-                user.fb_uid,
-                user.global_registry_mdm_id,
-            );
-        });
+        //Hack, we do not want the analytics to run when testing
+        if (!window.__karma__) analyticsService.init();
 
         $transitions.onBefore({}, transition => {
             if (transition.to().data && transition.to().data.isPublic)
--- a/app/assets/javascripts/angular/missionhubApp.run.js
+++ b/app/assets/javascripts/angular/missionhubApp.run.js
@@ -31,8 +31,14 @@ angular
         }
 
         facebookService.loadSDK()(document);
-        analyticsService.loadAdobeScript()(document);
-        analyticsService.init('SSOUID');
+
+        loggedInPerson.loadOnce().then(user => {
+            analyticsService.init(
+                user.thekey_uid,
+                user.fb_uid,
+                user.global_registry_mdm_id,
+            );
+        });
 
         $transitions.onBefore({}, transition => {
             if (transition.to().data && transition.to().data.isPublic)
--- a/app/assets/javascripts/angular/missionhubApp.run.js
+++ b/app/assets/javascripts/angular/missionhubApp.run.js
@@ -32,7 +32,7 @@ angular
 
         facebookService.loadSDK()(document);
 
-        loggedInPerson.loadOnce().then(user => {
+        loggedInPerson.load().then(user => {
             analyticsService.init(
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/angular/missionhubApp.run.js
```diff
@@ -31,8 +31,14 @@ angular
         }
 
         facebookService.loadSDK()(document);
-        analyticsService.loadAdobeScript()(document);
-        analyticsService.init('SSOUID');
+
+        loggedInPerson.loadOnce().then(user => {
```
OzzieOrca: Does this cause a duplicate load with line https://github.com/CruGlobal/missionhub-web/pull/87/files#diff-d879422f5b926cea61b6d56aab868c49R52 below? I know there's some sort of caching going on.

app/assets/javascripts/angular/services/analyticsService.js
```diff
@@ -31,41 +31,50 @@ function analyticsService($window, envService, $location) {
...
 
-                ref.parentNode.insertBefore(js, ref);
-            };
-        },
-        init: ssoUid => {
+            ref.parentNode.insertBefore(js, ref);
+        };
+    };
+
+    const setupAdobe = (ssoUid, facebookId, grMasterPersonId) => {
```
OzzieOrca: This doesn't use the params.

jesusfreak3: not sure i follow this comment

OzzieOrca:
```js
const setupAdobe = (ssoUid, facebookId, grMasterPersonId) => {
  $window._satellite && $window._satellite.pageBottom();
};
```
doesn't use `ssoUid, facebookId, grMasterPersonId`. They can be deleted here and where it gets called.

app/assets/javascripts/angular/missionhubApp.run.js
jesusfreak3: yes it will on initial run.  Both will fire.  was trying to figure out a way to do it once, but the second only only fires if they are not on a public route.  We need this to fire on initial run only and no matter if public route or not.

app/assets/javascripts/angular/services/analyticsService.js
jesusfreak3: oh sorry i see it, now, yes i refactored that and forgot to remove them.

app/assets/javascripts/angular/services/analyticsService.js
```diff
@@ -31,41 +31,54 @@ function analyticsService($window, envService, $location) {
...
+    };
+
+    const setupAdobe = () => {
+        $window._satellite && $window._satellite.pageBottom();
+    };
+
+    return {
+        init: () => {
+            loggedInPerson
+                .load()
```
OzzieOrca: I thought you were doing `loadOnce`...

jesusfreak3: Yah i was, i was testing things, and didnt put it back.  Technically it doesnt change to much since it will only ever run on a hard load.  But it would cache any next loadOnce calls.  In anycase I put it back to loadonce.

OzzieOrca: I think it'll help since we call it twice in quick succession in the run file (unless that is no longer the case) https://github.com/CruGlobal/missionhub-web/blob/master/app/assets/javascripts/angular/services/loggedInPerson.service.js#L59 should use the same promise both times. Idk if I have a full understanding though..

jesusfreak3: Yah so we call it in the before for authentication, but we do a new call each time in the before with no cache to check if the session is still good.  BUt there are other places in the App that do use loadonce as well, so this will precache that promise at least,

PR #88, opened 2018-12-20
Fix thekey redirect issues

jesusfreak3:


app/assets/javascripts/angular/missionhubApp.config.js
```diff
@@ -8,6 +8,7 @@ angular
                     'stage.missionhub.com',
                     'stage.mhub.cc',
                     'new.missionhub.com',
+                    '*.netlify.com',
```
OzzieOrca: Could you do `*-missionhub.netlify.com`? Makes it more specific. I guess this isn't really a security thing so maybe it doesn't matter. But would help you find which netlify subdomain we're using.

PR #89, opened 2018-12-20
remove duplicate breadcrumb from survey responses

jesusfreak3:


jesusfreak3:
@OzzieOrca lol looks like there was only spot, cant find any other occurrences of the breadcrumb.

PR #90, opened 2018-12-27
Mhp 2151 bearer token

jesusfreak3:
Ok moved the logic into the API error handling for catching 401s and other errors related to token issues.

### DIFF
--- a/app/assets/javascripts/angular/missionhubApp.run.js
+++ b/app/assets/javascripts/angular/missionhubApp.run.js
@@ -10,7 +10,6 @@ angular
         localStorageService,
         authenticationService,
         facebookService,
-        loggedInPerson,
         analyticsService,
     ) {
         lscache.setBucket('missionhub:');
@@ -43,13 +42,6 @@ angular
                 authenticationService.removeAccess();
                 return transition.router.stateService.target('signIn');
             }
-
-            return loggedInPerson.load().catch(e => {
-                if (e.status === 401) {
-                    authenticationService.removeAccess();
-                    return transition.router.stateService.target('signIn');
-                }
-            });
         });
 
         $transitions.onFinish({}, transition => {
--- a/app/assets/javascripts/angular/services/httpProxy.service.js
+++ b/app/assets/javascripts/angular/services/httpProxy.service.js
@@ -79,9 +79,6 @@ function proxyService(
                         }
                     }
 
-                    if (err.status === 401)
-                        $injector.get('authenticationService').removeAccess();
-
                     err.message = tFilter(config.errorMessage);
                     throw err;
                 });
--- a/app/assets/javascripts/angular/services/httpProxy.service.js
+++ b/app/assets/javascripts/angular/services/httpProxy.service.js
@@ -11,6 +11,7 @@ function proxyService(
     tFilter,
     errorService,
     _,
+    $injector,
 ) {
     // Extract and return the data portion of a JSON API payload
     function extractData(response) {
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/angular/services/httpProxy.service.js
```diff
@@ -53,7 +58,30 @@ function proxyService(
...
+                    ) {
+                        const errorDetail = data.errors[0].detail;
+
+                        if (
+                            EXPIRED_ACCESS_TOKEN === errorDetail ||
+                            INVALID_ACCESS_TOKEN === errorDetail ||
+                            INVALID_GRANT === errorDetail
+                        ) {
+                            $injector
+                                .get('authenticationService')
```
OzzieOrca: How come this can't be injected normally?

app/assets/javascripts/angular/services/httpProxy.service.js
```diff
@@ -53,7 +58,30 @@ function proxyService(
...
+                                .removeAccess();
+                        }
+                    }
+
+                    if (err.status === 401)
+                        $injector.get('authenticationService').removeAccess();
```
OzzieOrca: You can't do this. The API is stupid and throws a 401 when it should be throwing a 403. We'll just have to rely on the string errors above.

I asked for org `1` which I don't have access to (but I'm correctly signed in):
screen shot 2019-01-09 at 11 31 23 am

app/assets/javascripts/angular/missionhubApp.run.js
```diff
@@ -43,13 +42,6 @@ angular
                 authenticationService.removeAccess();
                 return transition.router.stateService.target('signIn');
             }
-
-            return loggedInPerson.load().catch(e => {
-                if (e.status === 401) {
-                    authenticationService.removeAccess();
-                    return transition.router.stateService.target('signIn');
```
OzzieOrca: What happened to this redirect to sign in? Does something else handle it now?

app/assets/javascripts/angular/services/httpProxy.service.js
jesusfreak3: thought you would ask about that, Angular throws a circular dependency and crashes out.  Happens since the authenticationService also calls on another service that this service calls.  So only way around is the injector.

app/assets/javascripts/angular/services/httpProxy.service.js
jesusfreak3: ahh right.  ill double check that.

jesusfreak3: I refactored it into the remove access from the service, https://github.com/CruGlobal/missionhub-web/blob/be301798691195a0a45e5e2c88e3a7ee3412e3f5/app/assets/javascripts/angular/services/authentication.service.js#L165 as anytime you remove the access you should be at sign in.

app/assets/javascripts/angular/missionhubApp.run.js
jesusfreak3: I refactoed it into the removeAccess from the authentication service. https://github.com/CruGlobal/missionhub-web/blob/be301798691195a0a45e5e2c88e3a7ee3412e3f5/app/assets/javascripts/angular/services/authentication.service.js#L165  
seemed like a better place since it should always go to signin if no access is set.

app/assets/javascripts/angular/services/httpProxy.service.js
OzzieOrca: Ah fun :) I like commenting when there are weird oddities like that. Helps people who discover it later.

app/assets/javascripts/angular/missionhubApp.run.js
OzzieOrca: Ok cool. I just was looking for that change in this PR and didn't see it.

app/assets/javascripts/angular/services/httpProxy.service.js
jesusfreak3: Ok removed the 401 catch since above i already added the cases by string errors.

PR #91, opened 2019-01-10
update to store new token and reuse if not passed

jesusfreak3:


app/assets/javascripts/angular/services/httpProxy.service.js
```diff
@@ -47,19 +48,26 @@ function proxyService(
...
+
                     if (token) {
+                        authenticationService.storeJwtToken(token);
                         $http.defaults.headers.common.Authorization =
                             'Bearer ' + token;
                     }
 
                     return JsonApiDataStore.store.syncWithMeta(res.data);
                 })
                 .catch(err => {
```
OzzieOrca: Could do
```suggestion
                .catch((err = {}) => {
```

app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -30,7 +31,7 @@ function authenticationService(
 
     const getJwtToken = () => {
         return (
-            (envService.is('development')
+            (envService.is('development') && !$window.__karma__
```
OzzieOrca: I'd like to get rid of `$window.__karma__` if possible. Would mocking sessionStorageService and localStorageService solve whatever this was needed for?

app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -55,21 +56,21 @@ function authenticationService(
...
 
-    const storeToken = token => {
+    const storeJwtToken = token => {
         if (!token) return;
 
         state.v4AccessToken = token;
         sessionStorageService.set('jwtToken', token);
 
-        if (envService.is('development'))
+        if (envService.is('development') && !$window.__karma__)
```
OzzieOrca: Same here

app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -140,7 +141,7 @@ function authenticationService(
 
         sessionStorageService.set('state', newState);
 
-        if (envService.is('development'))
+        if (envService.is('development') && !$window.__karma__)
```
OzzieOrca: Same here

app/assets/javascripts/angular/services/authentication.service.js
jesusfreak3: so to be honest, we can just remove it and it will work, it was more I only want to store to local storage if its the developer on his local, otherwise to use session storage.  So im just going to remove it.

jesusfreak3: hmm lost my comment.  SO we can just remove it, it wont break anything.  I just wanted to limit the localstorage of the authentication stuff to local dev env, but it doesnt really matter for tests.

PR #93, opened 2019-01-14
typeform integration

ardation:
This adds a modal explaining how to integrate Typeform to the surveys component.

There are new check marks showing if the form is from Typeform. If they click on the question mark it opens up the modal.
screen shot 2019-01-18 at 8 38 33 am

The button to access the modal is below the survey list
screen shot 2019-01-17 at 8 59 58 am


app/assets/javascripts/angular/components/surveyTypeformModal/surveyTypeformModal.html
```diff
@@ -0,0 +1,18 @@
+<div class="modal-header">
+    <h3 class="modal-title">{{ 'ministries.surveys.typeform.header' | t }}</h3>
+</div>
+<div class="modal-body form">
+    <ol>
+        <li bind-dynamic="'ministries.surveys.typeform.step1' | t"></li>
```
OzzieOrca: Idk that I've seen `bind-dynamic` before. Guess it's a custom directive. Would `ng-bind-html` or `ng-bind-html-unsafe` be better here? Doesn't seem like this needs to update ever.

app/assets/javascripts/angular/components/surveyTypeformModal/surveyTypeformModal.scss
```diff
@@ -0,0 +1,15 @@
+message-modal {
```
OzzieOrca: You don't seem to be using any of this css...

PR #94, opened 2019-01-15
Mhp 1670 ptc org signatures

jesusfreak3:


app/assets/javascripts/angular/components/organizationSignatures/organizationSignatures.component.js
```diff
@@ -0,0 +1,88 @@
...
+            return {
+                first_name: data.first_name,
+                last_name: data.last_name,
+                updated_at: data.updated_at,
+                code_of_conduct: data.code_of_conduct,
+                statement_of_faith: data.statement_of_faith,
+            };
+        });
+
+        return data;
```
OzzieOrca: I think this is dead code. There is a return above.

app/assets/javascripts/angular/components/organizationSignatures/organizationSignatures.component.js
```diff
@@ -0,0 +1,88 @@
...
+            ...(searchText ? { 'filters[q]': searchText } : {}),
+        };
+
+        return httpProxy.get(`/organizations/${orgId}/signatures`, params, {
+            errorMessage: 'error.messages.surveys.loadQuestions',
+        });
+    };
+
+    const prepareData = data => {
+        return data.map(data => {
```
OzzieOrca: I think reusing `data` is confusing. I know it's a trivial use case. `item` or `signature` might be better.

app/assets/javascripts/angular/components/organizationSignatures/organizationSignatures.component.js
```diff
@@ -0,0 +1,88 @@
...
+
+    const buildExportCsvLink = searchText => {
+        const params = {
+            access_token: $http.defaults.headers.common.Authorization.slice(7),
+            include: '',
+            format: 'csv',
+            ...(searchText ? { 'filters[q]': searchText } : {}),
+        };
+
+        const query = Object.keys(params).reduce((accumulator, key, index) => {
```
OzzieOrca:
```suggestion
        const query = Object.entries(params).reduce((accumulator, [key, value]) => {
```
works too. Then you don't have to manually index off of `params`.

app/assets/javascripts/angular/components/organizationSignatures/organizationSignatures.component.js
```diff
@@ -0,0 +1,88 @@
...
+        return (
+            envService.read('apiUrl') +
+            `/organizations/${this.orgId}/signatures?${query}`
+        );
+    };
+
+    const loadData = async searchText => {
+        const { data } = await getSignatures(this.orgId, searchText);
+        this.signatureData = prepareData(data);
+        this.signatures = [].concat(prepareData(data));
```
OzzieOrca: I'm a little confused why both `signatureData` and `signatures` exist. I get that it's something to do with the table. Calling `prepareData` twice might not be necessary.
```suggestion
        this.signatures = [...prepareData(data)];
```
would do the same thing I think if you like that any better.

app/assets/javascripts/angular/components/organizationSignatures/organizationSignatures.component.js
OzzieOrca: Or doing this might make it shorter:
```js
const prepareData = data =>
  data.map(
    ({
      first_name,
      last_name,
      updated_at,
      code_of_conduct,
      statement_of_faith
    }) => ({
      first_name,
      last_name,
      updated_at,
      code_of_conduct,
      statement_of_faith
    })
  );
```
Idk if that reads any better.

app/assets/javascripts/angular/components/organizationOverview/organizationOverview.html
```diff
@@ -51,6 +51,9 @@
                                 {{ 'nav.report_movement_indicators' | t }}
                             </a>
                         </li>
+                        <li>
```
OzzieOrca:
```suggestion
                        <li ng-if="$ctrl.p2cOrg">
```

app/assets/javascripts/angular/components/organizationOverview/organizationOverview.html
```diff
@@ -51,6 +51,9 @@
...
+                            <a ui-state="'^.ministry.signatures'" ui-state-params="{ orgId: $ctrl.org.id }" ui-sref-active="active">{{ 'signatures.title' | t }}</a>
+                        </li>
                         <li ng-if="$ctrl.p2cOrg">
```
OzzieOrca: This needs to be deleted.

app/assets/javascripts/angular/components/organizationSignatures/organizationSignatures.component.js
jesusfreak3: yah so it looks like they mutate the main array of data when sorting, and filtering etc, so the duplicate is their "referenced copy" they can always revert to etc.  They re built the way this table system works in new angualr, vue, and react, but in the angualr1 its still like this.

app/assets/javascripts/angular/components/organizationOverview/organizationOverview.html
jesusfreak3: yup, this was one of the things i was going to fix once i had all the data, since i was testing with non p2c originally.

app/assets/javascripts/angular/components/organizationOverview/organizationOverview.html
jesusfreak3: yup, same with this, i wanted to be able to test p2c logic first.

app/assets/javascripts/angular/components/organizationSignatures/organizationSignatures.component.js
jesusfreak3: hmm... yah refactored that part and forgot to remove the return data... we would not be a good God destroying so much of our creation and calling it useless! lol

app/assets/javascripts/angular/components/organizationSignatures/organizationSignatures.component.js
jesusfreak3: I adjusted slightly, not sure if the destructive version is easier on the eyes or not.

jesusfreak3:
@OzzieOrca thanks, so since i never had the proper data i was assuming declined and accepted results.  But now since i have the data I corrected declined, accepted and not yet accepted.  So just working on pagination now.

app/assets/javascripts/angular/missionhubApp.run.js
```diff
@@ -42,4 +42,14 @@ angular
                 });
             }
         });
+    })
+    .run(function($templateCache) {
+        $templateCache.put(
+            'template/smart-table/pagination.html',
```
OzzieOrca: You can use template literals to do multiline strings.

app/assets/javascripts/angular/components/organizationSignatures/organizationSignatures.component.js
OzzieOrca: [`@babel/plugin-transform-runtime`](https://babeljs.io/docs/en/babel-plugin-transform-runtime) just adds babel helper code.

I've got it configured to use [@babel/preset-env](https://babeljs.io/docs/en/babel-preset-env) which should by default make it work with these browsers https://browserl.ist/. Idk if that's the best configuration but seems sane enough.

app/assets/javascripts/angular/components/organizationSignatures/organizationSignatures.html
```diff
@@ -0,0 +1,75 @@
...
+                            </li>
+                            <li class="ml2" ng-repeat="page in $ctrl.tableState.pagination.pages" >
+                                <a href
+                                    class='f5 pv1 ph2 b--pivot-gray-darker pivot-white br2 dim'
+                                    ng-class="((page + 1) === $ctrl.tableState.pagination.currentPage) ? 'bg-pivot-gray-darker' : 'bg-pivot-gray-dark'"
+                                    ng-click="$ctrl.setPage(page, search);"
+                                    >{{ (page + 1) }}</a>
+                            </li>
+                            <li ng-if="$ctrl.tableState.pagination.currentPage < $ctrl.tableState.pagination.numberOfPages">
+                                <a href ng-click="$ctrl.nextPage(search);" class='ml1 f5 pv1 ph2 pivot-gray-darker br2 dim'> > </a>
```
OzzieOrca:
```suggestion
                                <a href ng-click="$ctrl.nextPage(search);" class='ml1 f5 pv1 ph2 pivot-gray-darker br2 dim'> &gt; </a>
```

app/assets/javascripts/angular/components/organizationSignatures/organizationSignatures.html
```diff
@@ -0,0 +1,75 @@
...
+                <td class='pa3 tc' colspan="5" ng-if="!$ctrl.isLoading">{{ 'signatures.grid.notFound' | t }}</td>
+            </tr>
+        </tbody>
+        <tfoot>
+			<tr>
+				<td colspan="5" class="text-center">
+                    <div class="pagination mt3" ng-if="$ctrl.tableState.pagination.pages.length > 1">
+                        <ul class="list flex">
+                            <li ng-if="$ctrl.tableState.pagination.currentPage > 1">
+                                <a href ng-click="$ctrl.previousPage(search);" class='f5 pv1 ph2 pivot-gray-darker br2 dim'> < </a>
```
OzzieOrca:
```suggestion
                                <a href ng-click="$ctrl.previousPage(search);" class='f5 pv1 ph2 pivot-gray-darker br2 dim'> &lt; </a>
```
I think this is killing Webpack's html-loader parser when running on Netlify.

app/assets/javascripts/angular/components/organizationSignatures/organizationSignatures.component.js
```diff
@@ -0,0 +1,128 @@
...
+    this.previousPage = searchText => {
+        this.tableState.pagination.start =
+            (this.tableState.pagination.currentPage - 2) * this.itemsPerPage;
+        loadData(searchText, this.tableState);
+    };
+
+    this.load = tableState => {
+        tableState.pagination.numberOfPages = 0;
+        this.tableState = tableState;
+        loadData(_, tableState);
```
OzzieOrca: `_` is probably undefined or the lodash object... Usually when you use `_` for a throwaway variable, you use it in an assignment. Am I missing something? Would one of these be better?
```suggestion
        loadData(undefined, tableState);
```
```suggestion
        loadData(null, tableState);
```

app/assets/javascripts/angular/missionhubApp.run.js
jesusfreak3: yah, meant to actually remove it lol.. was just there when I was testing it, converted it to literals and then remembered im not using it now.

app/assets/javascripts/angular/components/organizationSignatures/organizationSignatures.component.js
```diff
@@ -0,0 +1,128 @@
...
+    httpProxy,
+    $scope,
+    $window,
+    envService,
+    $http,
+) {
+    this.signatures = [];
+    this.signatureData = [];
+    this.exportLink = '';
+    this.itemsPerPage = 50;
```
OzzieOrca: If we lower this is the request any faster? `https://api-stage.missionhub.com/apis/v4/organizations/10129/signatures?page%5Blimit%5D=50` is taking 14 seconds for me. If a lower limit doesn't help (and even if it does) we should probably bring this up with an API dev. Have you been seeing this?

PR #96, opened 2019-01-17
Mhp 1668 signatures code and statement

jesusfreak3:


app/assets/javascripts/angular/components/organizationSignatures/organizationSignaturesSign.component.js
```diff
@@ -0,0 +1,86 @@
...
+            },
+        };
+
+        return httpProxy.post(`/organizations/${orgId}/signatures`, params, {
+            errorMessage: 'error.messages.surveys.loadQuestions',
+        });
+    };
+
+    const updateAgreement = (type, status) => {
+        state.organization_with_missing_signatures_ids.forEach(async orgId => {
```
OzzieOrca: Using `Promise.all` would be better so they can all fire at once
```js
await Promise.all(
  state.organization_with_missing_signatures_ids.map(orgId =>
    sendAgreement(orgId, type, status)
  )
);
```

app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -361,18 +366,6 @@ angular
                     ),
                 },
             })
-            .state({
-                name: 'app.ministries.ministry.import',
```
OzzieOrca: Why did this get deleted?

app/assets/javascripts/angular/components/organizationSignatures/organizationSignaturesSign.component.js
OzzieOrca: I guess your current code sorta does this but the inner async/await doesn't really do anything. `Promise.all` will let you wait until they all finish before continuing if that's what you want.

jesusfreak3: right, not sure if it actually matters, but have this work around as well.  I think i like it better then promise.all.

OzzieOrca: Can you describe your work around? I don't think it's doing anything... Making the `forEach` callback an `async` function just returns a promise, which `forEach` immediately ignores. The code after this is immediately run before any of the promises resolve...

OzzieOrca: Just noticed you updated the code...

app/assets/javascripts/angular/components/organizationSignatures/organizationSignaturesSign.component.js
```diff
@@ -0,0 +1,97 @@
...
+            },
+        };
+
+        return httpProxy.post(`/organizations/${orgId}/signatures`, params, {
+            errorMessage: 'error.messages.surveys.loadQuestions',
+        });
+    };
+
+    const updateAllAgreements = async (orgIds, type, status) => {
+        for (const orgId of orgIds) {
```
OzzieOrca: This runs all the requests sequentially. I think running them in parallel would be better. The only way I know of doing that is `Promise.all`.

OzzieOrca: Seems to at least be a good enough practice that eslint has a rule for it https://eslint.org/docs/rules/no-await-in-loop

jesusfreak3: yah, sorry i missread your first comment, i thought you were asking them to be sequential for some reason.. yah promise.all is the only way.. ill correct that.

OzzieOrca: Oh sorry if I wasn't clear. Thanks :)

app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -335,6 +335,23 @@ angular
                     },
                 },
             })
+            .state({
+                name: 'app.ministries.signAgreements',
+                url: '/sign-agreements',
+                component: 'organizationSignaturesSign',
+            })
+            .state({
+                name: 'app.ministries.ministry.import',
```
OzzieOrca: Looks like you moved this above the `app.ministries.ministry`. Idk if there is any reason to move it or not to move it.   Just want to make sure the order of the routes doesn't matter.

jesusfreak3: your right, i should be more carefull, order can matter, it wouldnt in this case, but order is used by the routerUI

PR #99, opened 2019-01-25
MHP-2333: Hide validate phone number feature unless org is typeform_e…

OzzieOrca:
…nabled

https://jira.cru.org/browse/MHP-2333

I'm a little confused about why this is needed. How do orgs become `typeform_enabled`? Is that something orgs can opt into? I thought they could just click the [add typeform button](https://github.com/CruGlobal/missionhub-web/pull/93/files#diff-add512720605f878ca581d5fe9738810R14) but I see now there's a check there too. Is that flag something that got manually turned on in the DB for NZ orgs?

Is there a reason to restrict these features to NZ? Seems like the fewer special case feature sets we add the better. Are there plans to roll it out globally? I never heard the motivation for adding phone validation. Typeform seems like it would be useful for everyone.

Reviewing this with whitespace off should be easier. Since Prettier doesn't format HTML files 😢 I just reformatted the whole file with Intellij.

### DIFF
--- a/app/assets/javascripts/angular/components/surveyOverviewSettingsTab/surveyOverviewSettings.html
+++ b/app/assets/javascripts/angular/components/surveyOverviewSettingsTab/surveyOverviewSettings.html
@@ -61,64 +61,69 @@
                         {{'surveys:settings.image_delete' | t}}
                     </button>
                 </div>
-                <div class="mt4 bb pb4 b--pivot-gray tl">
-                    <input
-                        type="checkbox"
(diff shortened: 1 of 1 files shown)


nateemerson:
@OzzieOrca at the moment this is a feature developed specifically for NZ ministry (both typeform integration and the phone number validation). We could potentially open this up to more...but that hasn't really been discussed..this functionality was something they needed that Tatai built specifically for them.

PR #100, opened 2019-01-25
MHP-2336: Fix survey question delete assigned person

jesusfreak3:
Found a bug on production when trying to test Alex issue. if you want to remove a contact from the rule, it wont work.  Then also saw that when you delete a question answer, it doesnt delete both rules only the one on screen.  Added logic and fixed that.

### DIFF
--- a/app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
+++ b/app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
@@ -391,7 +391,6 @@ function surveyOverviewQuestionsController(
         );
 
         if (ruleIdsToDelete.length > 0) {
-            console.log(ruleIdsToDelete);
             await Promise.all(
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -373,14 +373,33 @@ function surveyOverviewQuestionsController(
...
     this.deleteQuestionContent = async (question, answers, rule, index) => {
-        if (rule.id) {
-            await surveyService.deleteSurveyQuestionRule(
-                this.survey.id,
-                rule.id,
+        const keyword = rule.trigger_keywords;
+        const ruleIdsToDelete = question.question_rules.reduce(
+            (accumulator, r) => {
+                if (r.trigger_keywords === keyword && r.id) {
+                    return [...new Set(accumulator.concat(...[r.id]))];
```
```suggestion
                    return [...new Set([...accumulator, r.id])];
```

OzzieOrca: Does this mean that multiple question rules can have the same id? That seems weird.

PR #104, opened 2019-01-28
add logic to hide header and footer from router

jesusfreak3:


src/app.component.js
```diff
@@ -13,10 +13,12 @@ angular.module('missionhubApp').component('app', {
 function appController(periodService, $rootScope, state, analyticsService) {
     let deregisterEditOrganizationsEvent;
     let deregisterStateChangedEvent;
+    let deregisterUiStateChangedEvent;
 
     this.editOrganizations = false;
     this.getPeriod = periodService.getPeriod;
     this.currentOrganization = state.currentOrganization;
+    this.hideHeader = false;
```
OzzieOrca: Does hide footer get a default value? I guess it's undefined but consistency is good.

jesusfreak3: ahh sorry, i rushed that... i split up the logic after between headeer and footer and forgot to init it...

jesusfreak3:
@OzzieOrca So it looks like the part with 

`resolve: {
                    person: function(loggedInPerson) {
                        return loggedInPerson.loadOnce().then(user => {
                            if (user.beta_mode === null) {
                                $uibModal.open({
                                    component: 'newWelcomeModal',
                                    resolve: {},
                                    windowClass: 'pivot_theme',
                                    size: 'sm',
                                });
                            }
                        });
                    },
                },`

is causing the issue in the abstract.  If you remove this, then it works with the way you want it to work.  Looks like since we moved the load of app differently, then the authentication service is not fired at the same time so this one returns 401 and nothing catches it.

I pushed this to your attempt branch.

But this doesnt resolve everything. as now you have other issues, but i think we can resolve it, its the mperson dashbaord pulling loggedin.person, but thats not loaded now, as it used to be pre loaded with the beta check.  But I think I can get this to work quickly

PR #106, opened 2019-02-08
MHP-1814 Remove 'Revert to Legacy' Option on User Preferences

mikeallisonJS:
Remove associated code & unit test


OzzieOrca:
Let's just get rid of everything at once since I'm looking through it. Sorry I didn't scope this very well. Can you kill:

1. https://github.com/CruGlobal/missionhub-web/blob/226da0576117eec9e6ed355224de0d776c3ee0d4/app/assets/javascripts/angular/components/surveyOverview/surveyOverview.html#L16
2. My `<sunset-legacy-env-bar>` (part of my Netlify hack for legacy). Master will keep working since  it still will have that component and the corresponding Netlify config.
3. Make https://github.com/CruGlobal/missionhub-web/blob/818683df5e8661216f0ac6a0d7e1e0b9f9ba5ac8/app/assets/javascripts/angular/components/aboutModal/aboutModal.html#L8 point at https://get.missionhub.com/terms-of-service
4. Point https://github.com/CruGlobal/missionhub-web/blob/eab925628a9c668559b8e70f05884fdac8e459b5/app/assets/javascripts/angular/components/organizationContactImport/organizationContactImportStep1.html#L7 at the new create survey modal. Not sure if the modal is part of the router config. Just going to the org tab would probably be enough. Feel free to make this another Jira task as it is more of a functionality change than just deleting stuff.
5. https://github.com/CruGlobal/missionhub-web/blob/bd7dce3c08a1e65f762c86e373cc8e04d31cd52b/app/assets/javascripts/angular/components/peopleViewPerson/peopleViewPerson.html#L11 and https://github.com/CruGlobal/missionhub-web/blob/e8b9e425a1bdac4017acaf42454dde64c9a940c6/app/assets/javascripts/angular/components/unassignedPeople/unassignedPeople.html#L3 need to be updated. Probably could be a new ticket as idk that we support routing with filters to the contacts screen. Could be complex with how it's currently setup.

mikeallisonJS:
Looking at it, I'm guessing #4 & 5 should be new tickets. #4 seems simple enough, but also seems a bit off scope

app/assets/javascripts/angular/components/surveyOverview/surveyOverview.html
```diff
@@ -5,17 +5,14 @@
                 <img class="db mr1" ng-src="{{$ctrl.chevronLeftIcon}}"/>
             </a>
             {{$ctrl.survey.title}}
+        </div>
```
OzzieOrca: Is this in the right place? I think this is what's making the header look like
screen shot 2019-02-13 at 11 52 59 pm
instead of
screen shot 2019-02-13 at 11 52 53 pm

mikeallisonJS:
fixed bad end tag, should be good now

OzzieOrca:
Created
https://jira.cru.org/browse/MHP-2392
https://jira.cru.org/browse/MHP-2393

PR #108, opened 2019-02-11
Attempt improving app state route nesting

jesusfreak3:


jesusfreak3:
@OzzieOrca take a look, and let me know, then I can merge into the other MHP branch and update it to use a new abstract for brandless or what ever we call it.

app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -242,32 +242,18 @@ angular
                 name: 'app',
                 url: '',
                 abstract: true,
-                resolve: {
-                    person: function(loggedInPerson) {
-                        return loggedInPerson.loadOnce().then(user => {
-                            if (user.beta_mode === null) {
-                                $uibModal.open({
```
OzzieOrca: I asked Alex about this modal related to something Mike is working on. She wanted to keep it around in case some users still haven't seen it. I guess we'll remove it sometime after our new branch goes live. Not sure where it should go. I guess it can move to myPeopleDashboard too...

jesusfreak3: I can add this into the app run since it tehcnially only needs to occur once and first one.

jesusfreak3: I added this logic, but its in the v2 PR, we can delete this PR, and the initial one since the V2 is  merge of both of them.
https://github.com/CruGlobal/missionhub-web/pull/110
https://github.com/CruGlobal/missionhub-web/pull/104

PR #109, opened 2019-02-11
Mhp 1668 signatures code and statement - accepted and decline messages

jesusfreak3:
Logic added for showing accepted and declined messages.

### DIFF
--- a/app/assets/javascripts/angular/components/organizationSignatures/organizationSignaturesSign.html
+++ b/app/assets/javascripts/angular/components/organizationSignatures/organizationSignaturesSign.html
@@ -27,5 +27,6 @@
             <p>{{ 'signatures.agreements.declined.mainBlurb' | t }}</p>
             <p>{{ 'signatures.agreements.declined.requestAccessBlurb' | t }}</p>
         </div>
+        <button class='f5 ba pv2 ph4 b--pivot-dark-blue pivot-white br4 bg-pivot-dark-blue dim ml2' ui-sref="app.people">{{ 'signatures.agreements.declined.button' | t }}</button>
     </section>
 </div>
--- a/app/assets/javascripts/angular/components/organizationSignatures/organizationSignaturesSign.html
+++ b/app/assets/javascripts/angular/components/organizationSignatures/organizationSignaturesSign.html
@@ -1,26 +1,29 @@
 <div>
     <section ng-if="!$ctrl.hasSignedCodeOfConduct()">
         <h1>{{ 'signatures.agreements.codeOfConduct.title' | t }}</h1>
-
         <div bind-dynamic="'signatures.agreements.codeOfConduct.text' | t">
         </div>
-
         <div class='fr mt3'>
             <button class='f5 ba pv1 ph4 b--pivot-gray-darker pivot-white br4 bg-pivot-gray-darker dim' ng-click="$ctrl.declineAgreement('code_of_conduct')">Decline</button>
             <button class='f5 ba pv1 ph4 b--pivot-dark-blue pivot-white br4 bg-pivot-dark-blue dim ml2' ng-click="$ctrl.acceptAgreement('code_of_conduct')">Accept</button>
         </div>
-
     </section>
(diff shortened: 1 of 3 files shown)


src/i18n/locales/en-US.js
```diff
@@ -675,6 +675,15 @@ export default {
...
                 accept: 'Accept',
                 decline: 'Decline',
+                accepted: {
+                    message: 'Thank you! You may now use MissionHub',
+                    button: 'Continue to the MissionHub App',
+                },
+                declined: {
+                    title: 'You have declined!',
+                    message:
+                        '<p>Admins will need to accept the Code of Conduct and Statement of Faith before proceeding to MissionHub. Please talk this over with your local staff and student leadership.</p><p>You will need to request access again after having this discussion. </p>',
```
OzzieOrca: I'd prefer not to use HTML here...

app/assets/javascripts/angular/components/organizationSignatures/organizationSignaturesSign.html
```diff
@@ -1,26 +1,29 @@
...
             <button class='f5 ba pv1 ph4 b--pivot-gray-darker pivot-white br4 bg-pivot-gray-darker dim' ng-click="$ctrl.declineAgreement('statement_of_faith')">{{ 'signatures.agreements.decline' | t }}</button>
             <button class='f5 ba pv1 ph4 b--pivot-dark-blue pivot-white br4 bg-pivot-dark-blue dim ml2' ng-click="$ctrl.acceptAgreement('statement_of_faith')">{{ 'signatures.agreements.accept' | t }}</button>
         </div>
-
+    </section>
+    <section class='tc' ng-if="$ctrl.hasSignedCodeOfConduct() && $ctrl.hasSignedSignedStatementOfFaith() && !$ctrl.hasDeclined">
+        <h1>{{ 'signatures.agreements.accepted.message' | t }}</h1>
+        <button class='f5 ba pv2 ph4 b--pivot-dark-blue pivot-white br4 bg-pivot-dark-blue dim ml2' ui-sref="app.people">{{ 'signatures.agreements.accepted.button' | t }}</button>
+    </section>
+    <section class='tc' ng-if="$ctrl.hasDeclined">
```
OzzieOrca: I think this needs a continue button also. The current decision was to make the decline signature API request remove the user's org permission with the new org. They should be able to continue and see all the orgs they had before (but not the ones they declined since the API removed the org permission).

jesusfreak3: np.  I wasnt sure what permissions would be removed since technically when they sign, they are signing for any org that was not signed.

src/i18n/locales/en-US.js
jesusfreak3: fixed

app/assets/javascripts/angular/components/organizationSignatures/organizationSignaturesSign.html
jesusfreak3: ok added button

src/i18n/locales/en-US.js
```diff
@@ -675,6 +675,18 @@ export default {
...
+                    message: 'Thank you! You may now use MissionHub',
+                    button: 'Continue to the MissionHub App',
+                },
+                declined: {
+                    title: 'You have declined!',
+                    mainBlurb:
+                        'Admins will need to accept the Code of Conduct and Statement of Faith before proceeding to MissionHub. Please talk this over with your local staff and student leadership.',
+                    reqestAccessBlurb:
+                        'You will need to request access again after having this discussion.',
+                    button: 'Continue to the MissionHub App',
```
OzzieOrca: This is duplicated twice now. Maybe you could move it up a level.

jesusfreak3: Yah wasnt sure if Alex wanted a different message for this, you think it would be the same in the end?

OzzieOrca: Ya idk. I suppose it could say something like continue without joining new orgs or continue to existing orgs or something. You could ask her.

jesusfreak3: I updated it so its not duplicated, i assume it will be the same

PR #111, opened 2019-02-12
Mhp 2165 migrate request access form

jesusfreak3:


src/components/authentication/requestAccess.component.js
```diff
@@ -0,0 +1,50 @@
+import template from './requestAccess.html';
+//import './requestAccess.scss';
```
OzzieOrca: ?

src/components/authentication/requestAccess.component.js
```diff
@@ -0,0 +1,50 @@
...
+            type: 'access_request',
+            attributes: {
+                first_name: requestData.firstName,
+                last_name: requestData.lastName,
+                email: requestData.email,
+                org_name: requestData.organizationName,
+            },
+        };
+
+        const { data } = await httpProxy.post(
```
OzzieOrca: You using `data` for anything?

src/components/navigation/navHeader.html
```diff
@@ -60,10 +60,10 @@
             <a href="http://help.missionhub.com" class='pivot-white f4 pr3 pl3' target='_blank'>{{ 'nav.support' | t }}</a>
         </li>
         <li class='fl'>
-            <a href class='pivot-white f4 pr3 pl3' ui-sref="app.login">{{ 'nav.login' | t }}</a>
+            <a href class='pivot-white f4 pr3 pl3' ui-sref="signIn">{{ 'nav.login' | t }}</a>
         </li>
         <li class='fl' >
-            <a href class='pivot-white f4 pr3 pl3' href="/request_access">{{ 'nav.requestAccess' | t }}</a><!-- TODO: fix url when request access is migrated -->
+            <a href class='pivot-white f4 pr3 pl3' ui-sref="requestAccess">{{ 'nav.requestAccess' | t }}</a><!-- TODO: fix url when request access is migrated -->
```
OzzieOrca: Can you remove the comment after this since this PR should be the fix for it?

src/components/authentication/requestAccess.html
```diff
@@ -0,0 +1,62 @@
...
+                </div>
+            </form>
+        </div>
+    </div>
+    <div class='tc' ng-if="$ctrl.formSubmitted">
+        <h1 class='f2 ttu'>{{ 'requestAccess.success.thankYou' | t }}</h1>
+        <p class='f3'>{{ 'requestAccess.success.message' | t }}</p>
+        <p class="">
+            <a href="https://www.cru.org/digitalministry/apps-tools.html">
+                <img alt="" src="https://d2zuyxs7furuq7.cloudfront.net/assets/welcome/icon-1@2x-f2e30bdf09508898dcae33eb33517161.png">
```
OzzieOrca: Can you move these assets into this repo? `d2zuyxs7furuq7.cloudfront.net` is a "Cache for stage.missionhub.com Rails app" which will probably be deleted sometime after legacy is killed. I think the source files are in here https://github.com/CruGlobal/missionhub/tree/master/app/assets/images/welcome.

src/i18n/locales/en-US.js
```diff
@@ -422,6 +423,25 @@ export default {
             login_with_key: 'Log in with The Key',
             login_with_facebook: 'Log in with Facebook',
         },
+        requestAccess: {
+            title: 'Request Access',
+            description:
+                '<p>MissionHub was created by Cru to help non-profit ministries stay connected with all of the leaders, members and outreach within their movement.</p> <p>Request access and our spectacular support team will help you get started.</p>',
```
OzzieOrca: I think I'd prefer not to use html in translations/use bind dynamic if it's easy not to. I think you could just split this into 2 translations.

src/components/authentication/requestAccess.html
```diff
@@ -0,0 +1,62 @@
...
+                        <label class="db fw4 lh-copy f6">{{ 'requestAccess.form.lastName' | t }}</label>
+                        <input class="pa1 input-reset ba bg-transparent w-100" type="text" required ng-model="requestData.lastName">
+                    </div>
+                    <div class="mt3">
+                        <label class="db fw4 lh-copy f6">{{ 'requestAccess.form.organizationName' | t }}</label>
+                        <input class="pa1 input-reset ba bg-transparent w-100" type="text" required ng-model="requestData.organizationName">
+                    </div>
+                    <div class="mt3">
+                        <label class="db fw4 lh-copy f6">{{ 'requestAccess.form.email' | t }}</label>
+                        <input class="pa1 input-reset ba bg-transparent w-100" type="email" required ng-model="requestData.email">
```
OzzieOrca: The email input seems to have a darker underline... I didn't really see why that would be the case looking through the Chrome inspector...
screen shot 2019-02-12 at 11 25 53 am

src/components/authentication/requestAccess.component.js
```diff
@@ -0,0 +1,50 @@
...
+            {
+                data: params,
+            },
+            {
+                errorMessage: 'error.messages.requestAccess',
+            },
+        );
+
+        this.formSubmitted = true;
+        this.formSubmitting = false;
```
OzzieOrca: You aren't handling errors. If I type an email of `asdf@asdf` I get a 422 from the API and can never submit the form again until refreshing the page.

src/i18n/locales/en-US.js
jesusfreak3: Sure, any reason just so i can understand?  worried its to dangerous?

src/components/navigation/navHeader.html
jesusfreak3: yah :s blah need to pick up on those things...

src/components/authentication/requestAccess.component.js
jesusfreak3: not anymore....

src/components/authentication/requestAccess.html
jesusfreak3: yah sorry saw that, meant to dig into it, its not a class as they are all the same, so some runaway css inherited somewhere.

src/components/authentication/requestAccess.component.js
jesusfreak3: are you manually doing that as the browser stops you from doing that with the html5 forms with required?

jesusfreak3: anyways i will add try catch, not sure why i didnt do it

src/components/authentication/requestAccess.html
jesusfreak3: added extra pattern validation, and also added try catch so that a user can resubmit if there is an error

src/components/authentication/requestAccess.component.js
jesusfreak3: added extra pattern validation, and also added try catch so that a user can resubmit if there is an error

src/components/authentication/requestAccess.html
jesusfreak3: sorry yah i meant to.  You actually just need to place them into the component dir, anyways its done now.

jesusfreak3:
@OzzieOrca ok should be good for you to look at now.

src/components/authentication/requestAccess.html
```diff
@@ -23,7 +24,7 @@ <h1 class="f2 pivot-dark-blue fw6 mb4 mt0">
                     </div>
                     <div class="mt3">
                         <label class="db fw4 lh-copy f6">{{ 'requestAccess.form.email' | t }}</label>
-                        <input class="pa1 input-reset ba bg-transparent w-100" type="email" required ng-model="requestData.email">
+                        <input class="pa1 input-reset ba bg-transparent w-100" type="email" pattern="[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,63}$" required ng-model="requestData.email" >
```
OzzieOrca: https://github.com/angular/angular.js/blob/master/src/ng/directive/input.js#L27 is what Angular is using. I feel like this issue is something that always comes up in projects :) It must support an intranet email server or something like `my-email@mycompany`. I'd say it's fine to just let the server throw the error if it wants it stricter. But if you like your regex or want to ammend Angular's we could do that to give them better feedback on the client. I was just using `asdf@asdf` to see what happened in the case of a server error :)

jesusfreak3: Yah exactly, its like that as sdfs@asdfsd is a valid email address or sddsf@[127.0.0.1] etc..  yah i think keeping the pattern is fine, prevents unneeded trip to server if someone accidentally forgets the domain extension.

PR #113, opened 2019-02-15
Mhp 2134 migrate invite flow

jesusfreak3:


app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -112,7 +112,11 @@ function authenticationService(
             const response = await requestTicket(accessToken);
             const { data } = await requestV4Token(response.data.ticket);
             setAuthorizationAndState(data.token, data.recent_organization);
-            $state.go('app.people');
+            const inviteState = sessionStorageService.get('inviteState');
+
+            if (inviteState)
+                $state.go('appWithoutMenus.inviteLink', inviteState);
+            else $state.go('app.people');
```
OzzieOrca: Looks weird here too. Ternary probably looks better.

app/assets/javascripts/angular/services/sessionStorage.service.js
```diff
@@ -10,8 +10,8 @@ angular.module('missionhubApp').service('sessionStorageService', () => {
         destroy: key => {
             sessionStorage.removeItem(key);
         },
-        clear: () => {
-            sessionStorage.clear();
+        clear: key => {
+            sessionStorage.removeItem(key);
```
OzzieOrca: This now does the same thing as `destroy`. I noticed some place where you were calling the old `clear` with a key which it would ignore and just clear everything. Idk if that usage is still around.

src/components/authentication/inviteLink.component.js
```diff
@@ -0,0 +1,76 @@
...
+                    {},
+                    {
+                        errorMessage:
+                            'error.messages.inviteLink.loadRemeberToken',
+                    },
+                );
+
+                const me = await loggedInPerson.load();
+
+                if (parseInt(me.user.id, 0) === parseInt(data.user_id, 0)) {
```
OzzieOrca: What's the `0` for?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Description seems to say `0` and `undefined` behave the same.

I think you can omit it or specify a radix of `10`.

src/components/authentication/inviteLink.component.js
```diff
@@ -0,0 +1,76 @@
...
+                    sessionStorageService.clear('inviteState');
+                    $location.url(`/ministries/${this.orgId}/suborgs`);
```
OzzieOrca: Could you do this with a `$state.go`? Idk that it matters...

src/components/authentication/inviteLink.component.js
OzzieOrca: Actually to be consistent maybe parse these as strings instead... Ids are strings everywhere else in the API. We really need to make a ticket to make all the auth ids strings...

app/assets/javascripts/angular/services/sessionStorage.service.js
jesusfreak3: I think this is backwards?  I added the key to be able to use removeItem with a key, as it was backwards before?

app/assets/javascripts/angular/services/authentication.service.js
jesusfreak3: yah np, swithcing

app/assets/javascripts/angular/services/sessionStorage.service.js
jesusfreak3: Right  i see, let me double check,

jesusfreak3: Ok i cleared up the confusion.

src/components/authentication/inviteLink.component.js
jesusfreak3: So with out the second parameter, the linter/compiler complains, which is why i added the 0.

jesusfreak3: But I added compare by string

src/components/authentication/inviteLink.component.js
jesusfreak3: I could not unless i get the org object and pass it to the component.

PR #114, opened 2019-02-15
load adobe dev js script for staging enb

jesusfreak3:


app/assets/javascripts/angular/services/analyticsService.js
```diff
@@ -100,6 +95,21 @@ function analyticsService(
...
+        const url = envService.is('production')
+            ? '//assets.adobedtm.com/3202ba9b02b459ee20779cfcd8e79eaf266be170/satelliteLib-b704a4f0b9d6babb4eac8ccc7c8a4fbf9e33f0fb.js'
+            : '//assets.adobedtm.com/3202ba9b02b459ee20779cfcd8e79eaf266be170/satelliteLib-b704a4f0b9d6babb4eac8ccc7c8a4fbf9e33f0fb-staging.js';
+
+        return loadScript(url, 'adobe-analytics');
+    };
+
+    const loadAdobeDevScript = () => {
+        return loadScript(
+            '//assets.adobedtm.com/launch-ENe4ca7f50fed34edd995d7c6294e6b509-development.min.js',
```
OzzieOrca: I think this is supposed to replace the satelliteLib staging script. I'm not familiar with it. You could ask Clark or @mikeallisonJS as I think he's used it in another project.

jesusfreak3: thats what I thought too, but when i looked at the source code of that script and the other, they were not similar, one was the API, and the other did more of manipulating the statalite data etc.  
But ill double check with them

app/assets/javascripts/angular/services/analyticsService.js
```diff
@@ -98,18 +98,11 @@ function analyticsService(
     const loadAdobeScript = () => {
         const url = envService.is('production')
             ? '//assets.adobedtm.com/3202ba9b02b459ee20779cfcd8e79eaf266be170/satelliteLib-b704a4f0b9d6babb4eac8ccc7c8a4fbf9e33f0fb.js'
```
OzzieOrca: Just be aware that the request to update this to launch should also be coming. You could do it now if the script is available since sunset-legacy isn't on prod anywhere yet. Idk if it's a matter of just deleting `-development` off the url or not.

PR #116, opened 2019-02-20
Mhp 2183 impersonate user

jesusfreak3:


app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -287,6 +287,20 @@ angular
                     isPublic: true,
                 },
             })
+            .state({
+                name: 'app.impersonatePerson',
+                url: '/impersonate-person/:userId',
```
OzzieOrca: Let's make this `impersonate-user` so the route and the type of id you're entering match.

src/components/authentication/impersonatePerson.component.js
```diff
@@ -0,0 +1,23 @@
...
+    },
+});
+
+function impersonatePersonController(authenticationService, $state, $scope) {
+    this.$onInit = async () => {
+        if (!authenticationService.isTokenValid()) $state.go('app.people');
+
+        if (this.userId)
+            await authenticationService.impersonatePerson(this.userId);
+        else await authenticationService.stopImpersonatingPerson();
```
OzzieOrca: Can you use curlies here? Or ternary (might have to disable an ESLint rule). Just looks weird to me. Trivial comment :)

app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -172,6 +171,28 @@ function authenticationService(
         $rootScope.$broadcast('state:changed', state);
     };
 
+    const impersonatePerson = async userId => {
+        try {
+            const { data } = await $http.get(
+                `${envService.read('apiUrl')}/impersonations/${userId}`,
+            );
+            setAuthorizationAndState(data.token, data.recent_organization);
```
OzzieOrca: Do we need to refresh the app after this? JsonApiDataStore will have all kinds of invalid data cached. Maybe there's a way to purge it but we'd need to make sure all the Angular run stuff gets executed to load all the needed data.

app/assets/javascripts/angular/missionhubApp.routes.js
OzzieOrca: While you're at it, want to make all of your `impersonatePerson` states/functions also say user?

app/assets/javascripts/angular/services/authentication.service.js
jesusfreak3: Lol i knew you would say that, was debating just doing it:P  I hads it as impersonate person , as the api says the persons user id to impoersonate.. so i put everything as user id, but left it as person.  But had a short though... hes soo going to say just change it to user...

src/components/authentication/impersonatePerson.component.js
jesusfreak3: That did end up looking very weird after the prettier did its work.

app/assets/javascripts/angular/missionhubApp.routes.js
jesusfreak3: Lol i knew you would say that, was debating just doing it:P I hads it as impersonate person , as the api says the persons user id to impoersonate.. so i put everything as user id, but left it as person. But had a short though... hes soo going to say just change it to user...

app/assets/javascripts/angular/services/authentication.service.js
jesusfreak3: Nope, authorization service handles it, updates  state and sends events that user has changed.

jesusfreak3:
@OzzieOrca  ok should be good now.

app/assets/javascripts/angular/services/authentication.service.js
OzzieOrca: Ok auth should be good. Does `updateUserData` need to be called? I was just comparing it to what was in missionhub.run.js.

I'm still a little worried about the JsonApiDataStore cache. Should be able to do https://github.com/beauby/jsonapi-datastore/blob/master/DOCUMENTATION.md#reset. I think we want the experience of the admin to match the experience of the user as closely as possible. There could be some bugs where data isn't loaded but cuz the admin was browsing around first that data is preloaded and cached. Or could end up seeing data you don't have access to either going from admin to impersonating or ending impersonation and going back to your account.

app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -288,18 +288,18 @@ angular
...
-                component: 'impersonatePerson',
+                component: 'impersonateUser',
                 resolve: {
                     userId: ($state, $transition$) =>
                         $transition$.params().userId,
                 },
             })
             .state({
-                name: 'app.stopeImpersonatingPerson',
+                name: 'app.stopeImpersonatingUser',
```
OzzieOrca: typo

app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -287,6 +287,20 @@ angular
                     isPublic: true,
                 },
             })
+            .state({
+                name: 'app.impersonateUser',
+                url: '/impersonate-person/:userId',
```
OzzieOrca: I think you missed the one I really wanted changed lol

app/assets/javascripts/angular/services/authentication.service.js
jesusfreak3: Yah so updateUserData updates the state in the sessions/storage and also broadcasts an event for any state changes.  It also calls a load on the current user to reload the cached jsonAPI of the user and organizations immediately after the token and http headers are updated.  But I guess there are other places that we could be worried about?

OzzieOrca: I guess manually navigating to the `impersonate-user` does do a page refresh before starting... Idk if that's enough.

app/assets/javascripts/angular/missionhubApp.routes.js
jesusfreak3: lol... please shoot me....

app/assets/javascripts/angular/missionhubApp.routes.js
jesusfreak3: fixed

app/assets/javascripts/angular/missionhubApp.routes.js
jesusfreak3: fixed

app/assets/javascripts/angular/services/authentication.service.js
OzzieOrca: Ignore the `updateUserData` comment. `updateUserData` and `setupUserSettings` seem to do almost the same thing.

jesusfreak3: ok added the reset for jsonapi.

PR #117, opened 2019-02-21
Mhp 2392

jesusfreak3:


app/assets/javascripts/angular/components/organizationContactImport/organizationContactImportStep1.component.js
```diff
@@ -69,4 +73,20 @@ function organizationContactImportStep1Controller($scope) {
...
+                component: 'createSurvey',
+                resolve: {
+                    organizationId: () => this.org.id,
+                },
+                windowClass: 'pivot_theme',
+                size: 'sm',
+            })
+            .result.then(newSurvey => {
+                this.selectedSurvey = newSurvey;
+                $uibModal.close({});
```
OzzieOrca: Maybe I'm missing something but it seems like the modal closes itself: https://github.com/CruGlobal/missionhub-web/blob/sunset-legacy/app/assets/javascripts/angular/components/createSurvey/createSurvey.component.js#L42

jesusfreak3: your right, didnt catch that!

PR #118, opened 2019-02-21
Mhp 2393 link people dashboard counts to people tab

jesusfreak3:


app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -442,6 +442,53 @@ angular
...
                 ministryViewTabs.map(function(tab) {
+                    const formatArrayForFilter = filter => {
+                        if (!filter) return {};
+
+                        return filter
+                            .split(',')
+                            .map(value => {
+                                return { [value]: true };
+                            })
+                            .reduce((carry, f) => {
```
OzzieOrca: I think you can just do this in one:
```js
filter
  .split(",")
  .reduce((acc, value) => 
    ({ ...acc, [value]: true }))
```

OzzieOrca:
Also the second top bar is missing
screen shot 2019-02-21 at 4 17 20 pm
screen shot 2019-02-21 at 4 17 47 pm

app/assets/javascripts/angular/components/peopleFiltersPanel/peopleFiltersPanel.component.js
```diff
@@ -56,19 +78,6 @@ function peopleFiltersPanelController(
         sendFilters();
     };
 
-    this.resetFilters = () => {
```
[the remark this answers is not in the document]
jesusfreak3: thats my bad... ill fix that, i removed it when i was cleaning up, as i removed the logic in the controller, but then didnt double check the view.. will re add it back.

app/assets/javascripts/angular/missionhubApp.routes.js
jesusfreak3: it doesnt work, tried many ways, but never generated the return i desired.  In anycase I cleaned it up removed the returns etc so its less messy

app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -442,6 +442,53 @@ angular
...
+                                'organizationOverview' + _.upperFirst(tab),
+                            resolve: {
+                                queryFilters: /* @ngInject */ (
+                                    $transition$,
+                                    $location,
+                                ) => {
+                                    const filters = $location.search();
+
+                                    return {
+                                        ...{
```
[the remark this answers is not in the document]
jesusfreak3: yah second one is what I want, your right, i was writing it backwards

jesusfreak3:
@OzzieOrca so that only shows if in your local/session storage state has currentOrganization set.  

image

Looks like its not related to this PR, it disapears on refresh, IE re init of the app.  Taking a look.

jesusfreak3:
Ok fixed that issue with the subnav not appearing after refresh, its good now.

jesusfreak3:
@OzzieOrca so looking at the unassignedPreople, it requires organization component, and in the html it has an ng if checking for a report property on the organization vm.

https://github.com/CruGlobal/missionhub-web/blob/6f019c016a5559d4ee356bebdc20eee43b80807a/app/assets/javascripts/angular/components/unassignedPeople/unassignedPeople.html#L1

However, if you look at the organization component it never creates that property.

https://github.com/CruGlobal/missionhub-web/blob/6f019c016a5559d4ee356bebdc20eee43b80807a/app/assets/javascripts/angular/components/organization/organization.component.js#L45

So it is impossible for it to show?

app/assets/javascripts/angular/missionhubApp.routes.js
OzzieOrca: Oh it's not working cuz you need an initial value. So
```js
.reduce((acc, value) => 
    ({ ...acc, [value]: true }), {})
```

If you don't specify one, it uses the first element of the array. Which does work fine with your code since after your map the first element is an object with the first key you want in it.

I just copy/pasted your code and didn't think about it :) Sorry.

OzzieOrca:
Maybe that's been broken for a while. I don't remember ever seeing it. I guess we can move forward and ask Alex about it...

app/assets/javascripts/angular/missionhubApp.routes.js
jesusfreak3: of course..... I should have picked that up.

PR #119, opened 2019-02-22
MHP-2384 fix navigation elements flashing

mikeallisonJS:
add fade animation
add authenticationService.loaded
make some position absolute to not overlap
fix broken class attribute

### DIFF
--- a/app/assets/javascripts/angular/services/authentication.service.js
+++ b/app/assets/javascripts/angular/services/authentication.service.js
@@ -29,7 +29,6 @@ function authenticationService(
     )}/login?response_type=token&scope=fullticket&client_id=${envService.read(
         'theKeyClientId',
     )}&redirect_uri=${redirectUrl}`;
-    let loaded = false;
 
     const getJwtToken = () => {
         return (
@@ -148,8 +147,6 @@ function authenticationService(
         state.loggedIn = true;
 
         $rootScope.$broadcast('state:changed', state);
-
-        loaded = true;
     };
 
     const setState = (organization, person) => {
@@ -177,7 +174,7 @@ function authenticationService(
         state.organization_with_missing_signatures_ids = null;
         state.loggedIn = false;
         JsonApiDataStore.store.reset();
-        loaded = false;
+
         $rootScope.$broadcast('state:changed', state);
     };
 
@@ -206,7 +203,6 @@ function authenticationService(
     return {
         authorizeAccess: authorizeAccess,
         authorizeFacebookAccess: authorizeFacebookAccess,
-        loaded: () => loaded,
         storeJwtToken: storeJwtToken,
         removeAccess: () => {
             clearToken();
--- a/app/assets/javascripts/angular/services/authentication.service.js
+++ b/app/assets/javascripts/angular/services/authentication.service.js
@@ -29,6 +29,7 @@ function authenticationService(
     )}/login?response_type=token&scope=fullticket&client_id=${envService.read(
         'theKeyClientId',
     )}&redirect_uri=${redirectUrl}`;
+    let loaded = false;
 
     const getJwtToken = () => {
         return (
@@ -147,6 +148,8 @@ function authenticationService(
         state.loggedIn = true;
 
         $rootScope.$broadcast('state:changed', state);
+
+        loaded = true;
     };
 
     const setState = (organization, person) => {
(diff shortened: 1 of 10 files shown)


src/components/navigation/navHeader.html
```diff
@@ -44,17 +44,17 @@
...
-    <ul class="list fr" id="primary_nav" ng-if="!$ctrl.state.hasMissionhubAccess && $ctrl.state.loggedIn" >
-        <li style="fl">
+    <ul class="list fr fade" id="primary_nav" ng-if="$ctrl.authenticationService.loaded && !$ctrl.state.hasMissionhubAccess && $ctrl.state.loggedIn" >
+        <li class="fl">
             <a href class='pivot-white f4 mr2' ng-click="$ctrl.logout()">{{ 'nav.logout' | t }}</a>
         </li>
     </ul>
 
-    <ul class="list fr" ng-if="!$ctrl.state.loggedIn">
+    <ul class="list fr fade" ng-if="$ctrl.authenticationService.loaded && !$ctrl.state.loggedIn">
```
jesusfreak3: Here I would change this ng if for this 
`ng-if="!$ctrl.state.loggedIn && !$ctrl.loggedInPerson.loadingPromise"`, that would completely remove the flashing of the unauthenticated menu now, both on full reload, and on logging in.

src/components/navigation/navHeader.component.js
```diff
@@ -13,6 +13,7 @@ function navHeaderController(
     envService,
     authenticationService,
 ) {
+    this.authenticationService = authenticationService;
```
jesusfreak3: This can be removed now, since you can use already injected state..

src/components/navigation/navHeader.scss
```diff
@@ -55,7 +55,8 @@ nav-header {
             margin-bottom: 0;
 
             > li {
-                position: relative;
+                position: absolute;
```
jesusfreak3: Sorry im not sure why this is better?  Can you attatch print screen of overlap?  Doing this will cause future responsive adaptations.

app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -147,6 +148,8 @@ function authenticationService(
         state.loggedIn = true;
```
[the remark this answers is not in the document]
mikealbert: I think you meant @mikeallisonJS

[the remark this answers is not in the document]
jesusfreak3: @mikealbert   so sorry!

src/components/navigation/navHeader.scss
mikeallisonJS: untitled3

mikeallisonJS: It's a brief blink when the ng-ifs digest

OzzieOrca:
The fade definitely makes it better. My thought was if we know we are coming in on an auth route, it might be best to never show the unauthenticated menu items. Idk how possible that is or what we could put in the menu instead. I guess it could be a blue blank bar but that could look weird too. This is a somewhat trivial issue so idk that we need to do tons of work here.

jesusfreak3:
@OzzieOrca @mikeallisonJS  yah I suggested code that should handle that, so we can combine that with the fade.

app/assets/javascripts/angular/services/authentication.service.js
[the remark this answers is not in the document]
mikeallisonJS: It was acting the same as loggedIn. However, that wasn't the intention. The goal was to have a separate property that handled the loading state. !loggedin & !loaded wouldn't display the unauthenticated items.

The fade was to mostly handle the nav items instantly popping in and out.

mikeallisonJS:
@jesusfreak3 using ctrl.loggedInPerson.loadingPromise doesn't seem to work like I would expect.

@OzzieOrca that's my goal with auth.loaded, to have a simple way to know if the auth testing is complete

src/components/navigation/navHeader.html
mikeallisonJS: this causes the un-auth to disappear on logout, and still flashes on login afterwards

jesusfreak3: actually it doesnt, if you combine with the change for appWithoutMenus it doesnt show.  Ill make this commit, take a look let me know, you can revert it after.

PR #120, opened 2019-02-25
MHP-1603 missing name in modal

mikeallisonJS:


mikeallisonJS:
I tried variations on nesting, but il8next ignored all nested attempts. I could have faked it with pluralization too, but that just had a smell.

OzzieOrca:
Alright. Don't worry about it.

PR #123, opened 2019-02-26
MHP-2348: update ministries overview UI with revised look

jesusfreak3:


jesusfreak3:
@OzzieOrca alex and the designer updated the designs again, so these are the latest.

PR #136, opened 2019-02-28
add get mission hub path and redirect users landing at mhub.cc at a n…

jesusfreak3:
this should work, I cant really test it, but it should work.

### DIFF
--- a/app/assets/javascripts/angular/missionhubApp.run.js
+++ b/app/assets/javascripts/angular/missionhubApp.run.js
@@ -41,7 +41,7 @@ angular
             if (transition.to().data && transition.to().data.isPublic)
                 return true;
 
-            if ($location.host() === 'mhub.cc')
+            if ($location.host().includes('mhub.cc'))
                 $window.location.href = envService.read('getMissionHub');
 
             if (!authenticationService.isTokenValid()) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/angular/missionhubApp.run.js
```diff
@@ -39,6 +41,9 @@ angular
             if (transition.to().data && transition.to().data.isPublic)
                 return true;
 
+            if ($location.host() === 'mhub.cc')
```
OzzieOrca: Does this cover `stage.hub.cc` also?

jesusfreak3: no, right now it only handles mhub.cc, if you want me to add stage.hub.cc as well, then I can add env variables for this, to catch it based on env let mek now

OzzieOrca: Yes I want it added. You could just check if the hostname contains `mhub.cc` here or check with a regex that it ends in that.

jesusfreak3: @OzzieOrca  ok just pushed i just used a simple includes on the string.

PR #138, opened 2019-02-28
MHP-2418: Fix undefined orgPermission when accessing archive_date

OzzieOrca:
Do you think this is a safe assumption? Not sure why the org permissions aren't loaded.

### DIFF
--- a/app/assets/javascripts/angular/components/peopleScreen/peopleScreen.component.js
+++ b/app/assets/javascripts/angular/components/peopleScreen/peopleScreen.component.js
@@ -112,7 +112,8 @@ function peopleScreenController(
                     this.org.id,
                 );
 
-                if (orgPermission.archive_date !== null) return 'Archived';
+                if (!orgPermission || orgPermission.archive_date !== null)
+                    return 'Archived';
 
                 return personService.getFollowupStatus(person, this.org.id);
             },


app/assets/javascripts/angular/components/peopleScreen/peopleScreen.component.js
```diff
@@ -112,7 +112,8 @@ function peopleScreenController(
                     this.org.id,
                 );
 
-                if (orgPermission.archive_date !== null) return 'Archived';
+                if (!orgPermission || orgPermission.archive_date !== null)
```
jesusfreak3: hmm not sure, maybe we should just do a &&  if org is set

orgPermission && orgPermission.archive_date !== null

jesusfreak3: But your right, im not sure why they would not have any orgPermission, unless they denied access?

OzzieOrca: It's probably cuz it's not loaded right...

The people who are missing the orgPermission seem to be the archived contacts.

jesusfreak3: ok cool

PR #140, opened 2019-02-28
adjust for mobile view

jesusfreak3:


src/app.scss
```diff
@@ -28,7 +32,7 @@ app {
         margin-left: auto;
         padding-left: 15px;
         padding-right: 15px;
-        min-width: 765px; //Currently below this width components mess up so locking here for now.
+        //min-width: 765px; //Currently below this width components mess up so locking here for now.
```
OzzieOrca: Does this start making other things look weird? Any way to change this just for this screen? A lot of other things look broken as they don't collapse well...

jesusfreak3: yah, it could.  Ill add an ovveride

PR #141, opened 2019-02-28
MHP-2410 karma dependency

mikeallisonJS:
update webpack/karma/karma-webpack
add node-noop to all asset imports which aren't testing related


karma.conf.js
```diff
@@ -30,6 +31,13 @@ const karmaWebpackConfig = {
                 : []),
         ],
     },
+    plugins: [
+        ...webpackConfig.plugins,
+        new webpack.NormalModuleReplacementPlugin(
+            /\.(gif|png|jpg|jpeg|(sa|sc|c)ss)$/,
```
OzzieOrca: I guess tests are passing but I think we have some svgs too.

mikeallisonJS: the culprit was scss, the others I just included from mpdx's build pipeline. svg could easily be added to speed it up a bit

PR #142, opened 2019-02-28
MHP-2415 remove unassigned component

mikeallisonJS:
remove usage
remove translation


app/assets/javascripts/angular/components/unassignedPeople/unassignedPeople.component.js
```diff
@@ -1,33 +0,0 @@
...
-    vm.setUnassignedVisible = setUnassignedVisible;
-
-    vm.$onInit = activate;
-
-    function getUnassignedVisibleKey() {
-        return ['unassignedVisible', vm.organization.org.id].join(':');
-    }
-
-    function activate() {
-        var val = lscache.get(getUnassignedVisibleKey());
```
OzzieOrca: Can you uninstall `lscache` too? I think this was the only place it was used. Let me know if I'm wrong. Thanks :)

mikeallisonJS: nice catch

PR #143, opened 2019-02-28
MHP-2417 reactive header

mikeallisonJS:
update headers to be reactive on smaller screens

fixes headers, but still has other component issues


jesusfreak3:
@OzzieOrca  @mikeallisonJS  found some issues when quickly going through.

image

jesusfreak3:
@mikeallisonJS @OzzieOrca also now on larger screens the sub menu is full screen, just put it in the "main-container" class it will resolve it.
image

mikeallisonJS:
@jesusfreak3 @OzzieOrca I have addressed all but the search box. I couldn't find the width for the expanded view (always read 12px in chrome)

OzzieOrca:
https://github.com/CruGlobal/missionhub-web/blob/sunset-legacy/src/components/navigation/navSearch.scss#L27

PR #149, opened 2019-03-01
MHP-2425: Add .well-known folder

OzzieOrca:
Test links:
https://deploy-preview-149--missionhub.netlify.com/.well-known/apple-app-site-association
https://deploy-preview-149--missionhub.netlify.com/.well-known/assetlinks.json

### DIFF
--- a/src/.well-known/apple-app-site-association
+++ b/src/.well-known/apple-app-site-association
@@ -1,0 +1,11 @@
+{
+    "applinks": {
+        "apps": [],
+        "details": [
+            {
+                "appID": "DQ48D9BF2V.com.missionhub",
+                "paths": [ "/l/*", "/auth*" ]
+            }
+        ]
+    }
+}
--- a/index.ejs
+++ b/index.ejs
@@ -6,6 +6,8 @@
     <base href="/">
 
     <title>MissionHub</title>
+    <meta name="apple-itunes-app" content="app-id=447869440">
+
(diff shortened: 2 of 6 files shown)


src/.well-known/apple-app-site-association
```diff
@@ -0,0 +1,11 @@
...
+    "applinks": {
+        "apps": [],
+        "details": [
+            {
+                "appID": "DQ48D9BF2V.com.missionhub",
+                "paths": [ "/l/*", "/auth*" ]
+            }
+        ]
+    }
+}
```
frett: for backwards compatibility we also need this file in the root directory. Older versions of iOS look in the root directory only

OzzieOrca: Good to know. I read https://developer.apple.com/library/archive/documentation/General/Conceptual/AppSearch/UniversalLinks.html which said:

> You can place the file at the root of your server or in the .well-known subdirectory.

Copying one directory is easier but I can do that if it's still needed.

frett: anything pre-iOS 9.3 won't look in `.well-known`.
https://developer.apple.com/library/archive/qa/qa1919/_index.html

I guess a better question is, do we still have any users using that old of a version of iOS

OzzieOrca: I don't have access to Adobe Analytics. I just added it...
https://deploy-preview-149--missionhub.netlify.com/apple-app-site-association

PR #152, opened 2019-03-02
Load countrily-data module async

OzzieOrca:


OzzieOrca:
Ya locally there's only english. The others are downloaded from OneSky during the Travis deploy. They should show up on staging and prod.

PR #155, opened 2019-03-05
Mhp 2098 use no response filter

jesusfreak3:


app/assets/javascripts/angular/components/peopleFiltersPanel/peopleFiltersPanel.html
```diff
@@ -126,17 +126,20 @@
                     <label class="filter-option flex-column" ng-if="question.kind === 'TextField'">
                         <select
                             class="w-100"
+                            ng-init="$ctrl.filters.answerMatchingOptions[question.id] = 'contains'"
                             ng-model="$ctrl.filters.answerMatchingOptions[question.id]"
-                            ng-change="$ctrl.updateFilters()"
-                            ng-init="$ctrl.filters.answerMatchingOptions[question.id] = 'contains'">
+                            ng-change="$ctrl.updateFilters($ctrl.updateFilters($ctrl.filters.answerMatchingOptions[question.id], question.id))">
```
OzzieOrca: Why is `$ctrl.updateFilters` called twice?

app/assets/javascripts/angular/components/peopleFiltersPanel/peopleFiltersPanel.component.js
```diff
@@ -83,7 +83,18 @@ function peopleFiltersPanelController(
...
 
-    this.updateFilters = () => {
+    this.updateFilters = (option, questionId) => {
+        const questionFilter =
+            option === 'no_response' || option === 'any_response' ? option : '';
+
+        if (questionId) {
+            const questions = this.filters.questions;
+            this.filters.questions = {
+                ...questions,
```
OzzieOrca: I don't think you need the `questions` variable.
```suggestion
                ...this.filters.questions,
```

app/assets/javascripts/angular/components/peopleFiltersPanel/peopleFiltersPanel.html
jesusfreak3: not sure, its how it has always been, they define a init, which i guess triggers the ng change?  But the ng initi is needed to set the default option for each question?

OzzieOrca: No, literally why is it `ng-change="$ctrl.updateFilters($ctrl.updateFilters(`? `updateFilters` is nested inside `updateFilters`.

app/assets/javascripts/angular/components/peopleFiltersPanel/peopleFiltersPanel.component.js
jesusfreak3: yup, true, should have cleaned that up, was attempting another way before this version.

app/assets/javascripts/angular/components/peopleFiltersPanel/peopleFiltersPanel.html
jesusfreak3: ahaha... oh... wow.. i didnt even see that.. thats incredible...

PR #156, opened 2019-03-05
Mhp 1630 my profile org id handling

jesusfreak3:


app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -197,12 +197,12 @@ angular
                         },
                     ),
 
-                    organizationId: lazyLoadedResolve(
-                        /* @ngInject */ function($uiRouter) {
-                            var $transition$ = $uiRouter.globals.transition;
-                            return $transition$.params().orgId;
-                        },
-                    ),
+                    organizationId: $uiRouter => {
```
OzzieOrca: I think you can inject `$transition$` directly. Maybe I'm wrong...

OzzieOrca: I can't seem to open  the modal with the netlify link. Maybe you still need the `/* @ngInject */`. It tells angular-annotate to wrap it in that array syntax so when the minifier renames the variables it doesn't break dependency injection.

jesusfreak3: > I think you can inject `$transition$` directly. Maybe I'm wrong...

I tried that, but when ever i tried it, the modal would never open, could not console to log either.  Wasnt sure why.  very weird behavoir

jesusfreak3: Ok pushed with nginject will see if it fixes it.

PR #157, opened 2019-03-05
Mhp 2420 cleanup boostrap

jesusfreak3:


jesusfreak3:
There are still some components being loaded, dropdown and modal and in return the bootstrap reboot as depended.  But I have removed the other components, the ones that clash with tachyon, so i think we are good now.  We could phase out the modal and the dropdown, but not sure if that is really needed.

PR #158, opened 2019-03-05
MHP-2439: update facebook to request email at login

jesusfreak3:


app/assets/javascripts/angular/services/facebookService.js
```diff
@@ -32,17 +32,20 @@ function facebookService(authenticationService, envService) {
...
+                        authenticationService.authorizeFacebookAccess(
+                            response.authResponse.accessToken,
+                        );
+                    } else {
+                        throw new Error(
+                            'User cancelled login or did not fully authorize.',
+                        );
+                    }
+                },
+                { scope: 'email' },
```
OzzieOrca: Is the server what uses this permission?

jesusfreak3: Yah exactly, we pass the facebook token to the server, and then they pull the data they need into the session etc.  BEfore we never had email, so they never had email on the server side, so merging current accounts failed since email was not set on that user etc..

PR #159, opened 2019-03-06
MHP-2383: Add locale and update my org dashboard list to show no ministries wer…

jesusfreak3:
…e found


app/assets/javascripts/angular/components/myOrganizationsDashboardList/myOrganizationsDashboardList.html
```diff
@@ -6,4 +6,7 @@
             load-details="false"
             edit-mode="$ctrl.myOrganizationsDashboard.editOrganizations">
     </organization-overview>
+    <section class='tc' ng-if="$ctrl.rootOrgs.length <=0">
```
OzzieOrca: I need to enable prettier on HTML now that it supports it. As long as it doesn't butcher the Angular template syntax...
```suggestion
    <section class='tc' ng-if="$ctrl.rootOrgs.length <= 0">
```

jesusfreak3: thx, yah, usually im pretty good with those spaces on conditions, but it slipped!

PR #160, opened 2019-03-07
MHP-2421 remove call to updateUserData of auth service on init after setupAUth…

jesusfreak3:
…enitcationState since same logic is called within


jesusfreak3:
So at one point i added that method in the init after the setupauthstate call.  It was related to siging agreements.  The setupAuthenticationState calls the same logic that the updateUserData did, so it should be good.  But that is why sometimes the avatar  doesnt show the correct user.  The load UI state is called 3 times with imerpsonate like that, first on init, but then a second in init asynchronously, and then third after impersonate is successful.  So the second call was occurring and sometimes occurring after impersonate was successful.

PR #161, opened 2019-03-07
Mhp 2440 update icons

jesusfreak3:


app/assets/javascripts/angular/components/iconButton/iconButton.component.js
```diff
@@ -21,6 +22,10 @@ function iconButtonController() {
 
     vm.handleClick = handleClick;
 
+    this.$onInit = () => {
+        this.size = this.size ? this.size : 24;
```
OzzieOrca:
```suggestion
        this.size = this.size || 24;
```
Would be the same thing. Also this is a one way input binding but this check is only done once... I don't think we'd ever update it though. Could add the `|| 24` to the template...

jesusfreak3: yah your right, simpler within the template.  No need for init.

PR #163, opened 2019-03-11
Mhp 2450 minor ui changes

jesusfreak3:


jesusfreak3:
> The right padding on the right icon of the blue row looks off.
> Screen Shot 2019-03-11 at 12 22 26 PM
> 
> Everything else looks great!

Ok fixed that part too, was the flex grow, never grew enough...

Going to add 2 other minor adjustments, so will let you know when to check code.

jesusfreak3:
@OzzieOrca  ok should be good now.

app/assets/javascripts/angular/components/ministryViewPerson/ministryViewPerson.html
```diff
@@ -4,7 +4,7 @@
             <input type="checkbox" ng-model="$ctrl.selected"/>
         </div>
         <div class="column name-column">
-            <a ui-sref=".person.defaultTab({ personId: $ctrl.person.id })">
+            <a href ui-sref=".person.defaultTab({ personId: $ctrl.person.id })">
```
OzzieOrca: Doesn't the `ui-sref` create a `href`? Whatever... Idk how it works anymore haha

app/assets/javascripts/angular/components/organization/organization.html
```diff
@@ -1,13 +1,13 @@
-<accordion collapsed="$ctrl.collapsed" collapsible="$ctrl.collapsible && !$ctrl.editMode">
+<accordion class='shadow-4' collapsed="$ctrl.collapsed" collapsible="$ctrl.collapsible && !$ctrl.editMode">
     <accordion-header>
-        <div class="header br2" ng-class="{'hidden-org': !$ctrl.org.visible}">
+        <div class="header" ng-class="{'hidden-org': !$ctrl.org.visible, 'br2' : $ctrl.collapsed !== false, 'br2 br--top' : $ctrl.collapsed === false}">
```
OzzieOrca: Your spacing before the colon is inconsistent. Also would `$ctrl.collapsed === true` read better than `$ctrl.collapsed !== false`? Idk if there are `undefined` or `null` cases this is handling weirdly.

app/assets/javascripts/angular/components/organizationOverview/organizationOverview.html
```diff
@@ -17,12 +17,12 @@
...
 
         <div class="tabs">
-            <div class="tab "
+            <div class="tab pivot-white"
                 ng-repeat="tabName in $ctrl.tabNames" ui-sref-active="active"
                 ui-state="'^.ministry.' + tabName" ui-state-params="{ orgId: $ctrl.org.id }"
                 ng-if="tabName !== 'labels'"
             >
-                <span class='pivot-white'>
+                <span class=''>
```
OzzieOrca: You can toss the `class`.

app/assets/javascripts/angular/components/ministryViewPerson/ministryViewPerson.html
jesusfreak3: it does, I have seen some weird behavoir, as many css is linked to link with href, so before the href exist those css styles could not be linked until the binding.

OzzieOrca: k

jesusfreak3: I know certain css frameworks used href on as to allow color adjustments, and also point changes (IE hand) , Bootstrap was one of them, which we still use for the reboot.  I see things like this

```
a:not([href]):not([tabindex]) {
    color: inherit;
    text-decoration: none;
```

etc..

OzzieOrca: Ya I'm fine with leaving it. Seems like the `ui-sref` would add it eventually but it might flash or something.

jesusfreak3: yah exactly.

PR #164, opened 2019-03-12
force the key access destroyed when invalid token present

jesusfreak3:


app/assets/javascripts/angular/services/httpProxy.service.js
```diff
@@ -85,7 +85,7 @@ function proxyService(
                             MISSING_TOKEN === errorDetail ||
                             INVALID_GRANT === errorDetail
                         ) {
-                            authenticationService.removeAccess();
+                            authenticationService.destroyTheKeyAccess();
```
OzzieOrca: Seems like logging out of The Key shouldn't be necessary but we could try it... My guess is this has something to do with session sharing between tabs.

OzzieOrca:
Duplicate of https://github.com/CruGlobal/missionhub-web/pull/182

PR #166, opened 2019-03-26
Mhp 2416 bring permission fitlers in contacts

jesusfreak3:


jesusfreak3:
> This needs to be added [here](https://github.com/CruGlobal/missionhub-web/blob/b4fa16e1358e816e551198aa0df87e8b82030f5b/app/assets/javascripts/angular/services/personSelection.js#L25) too so mass selection works with this new filter.

Done

PR #177, opened 2019-03-28
MHP-2539 update modal header for group members. fixed padding on icon button

jesusfreak3:


app/assets/javascripts/angular/components/groupMembersModal/groupMembersModal.html
```diff
@@ -1,6 +1,5 @@
 <div class="modal-header">
     <h3 class="modal-title">{{ 'groups.members.title' | t }}</h3>
-    <h4 class="modal-subtitle">{{ 'groups.members.subtitle' | t: { group_name: $ctrl.group.name } }}</h4>
```
OzzieOrca: How come we're removing the group name? Seems like it would help add context. I didn't look at the Jira issue yet.

jesusfreak3: It was requested by Eric in Figma

PR #178, opened 2019-03-28
MHP-2540 update dashboard sub nav to full expand and match main menu padding

jesusfreak3:


app/assets/javascripts/angular/components/dashboard/subnav/dashboard-subnav.scss
```diff
@@ -1,4 +1,8 @@
 dashboard-subnav {
     display: block;
     height: 100%;
+
+    .w-100 {
+        padding: 0 55px;
```
OzzieOrca: This feels like it should be another class. Adding padding functionality to a width class seems weird...

jesusfreak3: i know , i know...  but i know this wont break, as if they change from w-100 then they are adjusting the layout.. so its why i left it.

OzzieOrca: I'm confused and not sure I understand what would break. I'm less against it cuz it's at least scoped to the component. But seems unexpected that `w-100` would add padding.

jesusfreak3: Yah, there was no other placeholder to attach the custom padding to match the main menu padding above.  I could have added some random class to link it to, but decided just to do that instead of adding a new class and putting it there.  I was on the fence about it, but thought it should be ok since it wont and cant affect anything else

OzzieOrca: Just add a new class :)

jesusfreak3: Fine fine :P  just couldnt think of a name for a class before, but i think this will do

PR #184, opened 2019-04-19
MHP-1896: Hiding Surveys tab for non-direct-admin users

bwtoptal:


app/assets/javascripts/angular/components/organizationOverview/organizationOverview.html
```diff
@@ -45,7 +45,7 @@
                 ui-sref-active="active"
                 ui-state="'^.ministry.' + tabName"
                 ui-state-params="{ orgId: $ctrl.org.id }"
-                ng-if="tabName !== 'labels'"
+                ng-if="tabName !== 'labels' && $ctrl.isAllowed(tabName)"
```
OzzieOrca: This seems like a weird split of business logic. `tabName !== 'labels'` is in the template while `!(tabName === 'surveys' && !this.directAdminPrivileges)` is in the controller. My suggestion would be to move them both to the controller.

app/assets/javascripts/angular/components/organizationOverview/organizationOverview.component.js
```diff
@@ -33,6 +33,10 @@ function organizationOverviewController(
     this.toggleVisibility = userPreferencesService.toggleOrganizationVisibility;
     this.surveyResponses = 'countHidden';
 
+    this.isAllowed = tabName => {
```
OzzieOrca: Being nitpicky but I don't really like the name `isAllowed`. Seems kinda vague as to what is allowed and for whom. With the addition of the other code maybe `showTab`, `shouldShowTab`, `isTabVisible` or something would be better. I guess there are 2 things going on: hiding the labels tab so it can be shown in the dropdown menu and hiding the surveys tab when a user doesn't have valid permissions.

PR #185, opened 2019-04-23
MHP-2599: Fixing status column font size

bwtoptal:


app/assets/javascripts/angular/components/ministryViewPerson/ministryViewPerson.scss
```diff
@@ -43,14 +43,28 @@ ministry-view-person {
         }
     }
     .status-column {
-        .warning .form-control-select::after {
+        display: flex;
+        height: 50px;
```
OzzieOrca: Is this hardcoded height needed?

PR #186, opened 2019-04-26
Add null check to http error response data key

OzzieOrca:
Check that http error response has a data key before parsing error


OzzieOrca:
I think it will show a generic error toast. This http thing is wrapped in some error service that does it. I think https://github.com/CruGlobal/missionhub-web/pull/186/files#diff-25fa08d3fa470ac2119161b43f13f9ecR93 passes on the user facing error message. It might have not done it before cuz of the error through. I think it should be good but we can revisit it.

sheldond:
Sounds good, thanks for checking that!

PR #187, opened 2019-05-07
MHP-2521: Remove request access tab

bwtoptal:


bwilczyn:
I wasn't sure either, I thought it's just about temporarily hiding a tab, but I clarified with Alex and removed the component completely.

PR #190, opened 2019-05-15
MHP-2601: Fix unarchive button formatting

bwtoptal:


bwtoptal:
Thanks for pointing this out! I did a quick fix.

PR #191, opened 2019-05-17
MHP-2614: Adding fields to organization with groups request

bwtoptal:


bwtoptal:
I added fields to request for a specific organization, so it shouldn't return any relationships of group entity. 

The second place you mentioned (https://github.com/CruGlobal/missionhub-web/blob/f84b4a0ff0064f8452ca16d0690ede52cec6de93/app/assets/javascripts/angular/services/groups.service.js#L245) sends a request to `groups` endpoint and from what I see in docs (https://api-stage.missionhub.com/docs/apis::v4::group/get_user's_groups) it doesn't allow to pass `fields` parameter (I tried to pass them anyway, but they only limit `attributes` and not `relationships`).

Also, I think there is one more place where `/apis/v4/organizations/xxxx` with params {"include"=>"labels,groups"} is called and it's here: 
https://github.com/CruGlobal/missionhub-web/blob/f84b4a0ff0064f8452ca16d0690ede52cec6de93/app/assets/javascripts/angular/services/routes.service.js#L40

so I also added `fields` there which looks like speed up the query from 600-700ms to 350-450ms based on my experiments.

app/assets/javascripts/angular/services/routes.service.js
```diff
@@ -42,6 +42,10 @@ function routesService(httpProxy, modelsService) {
                 {
                     errorMessage: 'error.messages.routes.get_organization',
                 },
+                {
+                    'fields[group]':
```
OzzieOrca: Seems like this wouldn't get any fields for the organization itself and only get group attributes. Maybe I have an incorrect understanding of how fields work. If you don't specify the top level, does it load all of them?

bwilczyn: Yes, I think this is how it works. Based on the spec here: https://api-stage.missionhub.com/docs/apis::v4::organization/get_user%27s_organizations if you don't specify `fields` restriction then all the fields will be returned for the given entity (the entity for which we specify `fields`). Also, I checked the requests and in fact, all the organizations' attributes are returned and only group entities are restricted to the specified fields. I just wanted to get rid of these circular dependencies.

app/assets/javascripts/angular/components/organizationOverview/organizationOverview.service.js
```diff
@@ -34,6 +34,8 @@ function organizationOverviewService(
                 modelsService.getModelUrl(org),
                 {
                     include: include.join(','),
+                    'fields[group]':
```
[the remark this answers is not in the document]
bwtoptal: You are right sometimes surveys are also loaded here (I think it checks if the surveys were already loaded and loads them only if they weren't). Since we didn't put `fields[survey]` here, it loads all the survey attributes as before my change. We only restrict group attributes to not follow circular dependencies.

PR #193, opened 2019-05-21
MHP-2634: Contact view on people screen

bwtoptal:


bwtoptal:
I don't have any organizations or people assigned on production. But when I check master branch locally and click on a row with people a modal is displayed and url is `https://localhost:8080/people/73614/profile`. 

So it's missing organization id part. On deployed branch, I'm getting url: `https://deploy-preview-193--missionhub.netlify.com/people/73614/organization/4521/profile` which looks good.

OzzieOrca:
Does it navigate wrongly all the time?

Seems to work here:
Screen Shot 2019-05-21 at 11 41 04 AM

Almost seems like you edited the wrong route unless it's shared. Clicking My Profile in the header dropdown menu takes me here with a weird url of `people/2186671/organization//profile`:
Screen Shot 2019-05-21 at 11 41 46 AM

Staging could be out of date if anyone was testing there... I just reset staging to master.

bwtoptal:
You're right.. Clicking your own profile shows weird URL, I'll fix that. However, this task was about showing "Contact View" with tabs when clicking on a row with a user, in people view. Or maybe I got it wrong?

OzzieOrca:
I was just confused cuz it seemed to be working for me from the contact tab. But obviously there's a bug somewhere if it's reproducible.

bwtoptal:
Hmm.. It does work on the contact tab. I think this task was about showing tabs in the modal on the people screen, so when you select People in the main navigation and click one of the users.

OzzieOrca:
Oh gotcha. Ya that makes more sense. Seems like your code fixes that. Not sure what can be done about the My Profile url. That one specifically doesn't have an org so it purposely doesn't show tabs. Guess that functionality got carried over to clicking people on the people dashboard.

OzzieOrca:
If there's not an easy solution, I can just approve this. The weird url isn't the end of the world. But if it's easy to fix we should do it.

bwtoptal:
It was quite easy to fix. I've just added a new state which is only used by My Profile link, so it doesn't need to have `organization` URL part. Thanks for pointing this issue out! 👍

app/assets/javascripts/angular/missionhubApp.routes.js
```diff
@@ -525,6 +525,16 @@ angular
             .states(
                 generatePersonPageStates({
                     name: 'app.people.person',
+                    url: '/:personId/organization/:orgId',
+                    modal: true,
+                    params: {
+                        orgId: { value: null },
```
OzzieOrca: What does this do? Just default it to null?

bwtoptal: Hmm.. I guess so.. I did exactly the same as above. `orgId` has to be there, otherwise, it won't show up as a param.

OzzieOrca: Ok. I was just wondering if it's necessary. If it's needed keep it :)

PR #194, opened 2019-05-30
MHP-2482: Initial React dashboard component

bwtoptal:


package.json
```diff
@@ -7,6 +7,7 @@
         "build": "webpack -p",
         "build:analyze": "webpack -p --env.analyze",
         "test": "karma start --single-run",
+        "test-react": "jest",
```
OzzieOrca: I've been using colons here (probably just cuz mobile started doing  that). But doesn't really matter. Here's some other scripts we've added https://github.com/CruGlobal/missionhub-react-native/blob/develop/package.json#L9-L12. Don't necessarily need them all now.

package.json
```diff
@@ -73,11 +91,16 @@
         "@babel/runtime": "^7.4.3",
         "@brainly/onesky-utils": "^1.3.1",
         "angular-mocks": "^1.7.8",
+        "apollo-link-mock": "^1.0.1",
```
OzzieOrca: I mentioned it in my slack message but I was playing with https://www.apollographql.com/docs/graphql-tools/mocking/ https://github.com/CruGlobal/missionhub-react-native/pull/985. Any thoughts on doing it that way? Idk what's best. Having to spell out all the data for every test seems tedious.

package.json
```diff
@@ -73,11 +91,16 @@
...
         "babel-eslint": "^10.0.1",
         "babel-loader": "^8.0.4",
         "babel-plugin-angularjs-annotate": "^0.10.0",
+        "babel-polyfill": "^6.26.0",
         "copy-webpack-plugin": "^5.0.2",
         "css-loader": "^2.1.1",
+        "enzyme": "^3.9.0",
```
OzzieOrca: Any thoughts on https://testing-library.com/docs/react-testing-library/intro?

src/dashboard/components/navigation.spec.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import { ApolloClient } from 'apollo-client';
+import { ApolloProvider } from 'react-apollo-hooks';
+import { InMemoryCache } from 'apollo-cache-inmemory';
+import { MockLink } from 'apollo-link-mock';
+import { GET_ORGANIZATIONS } from '../graphql';
+import Navigation from './navigation';
+
+function createClient(mocks) {
```
OzzieOrca: These testing helpers should probably go in a dedicated file eventually. They can stay for now since you're only testing one component.

src/dashboard/components/navigation.spec.js
```diff
@@ -0,0 +1,50 @@
...
+                            name: 'Test Organization',
+                            id: '1',
+                        },
+                    },
+                },
+            },
+        ];
+
+        const wrapper = mount(
+            <ApolloProvider client={createClient(mocks)}>
```
OzzieOrca: Depending on the testing library you could create a helper render function like this: https://github.com/CruGlobal/missionhub-react-native/blob/42a63e8b0a7b63601a89c9b06d0837d23e51be3d/testUtils/index.tsx#L29-L60

src/dashboard/graphql/getCelebrations.js
```diff
@@ -0,0 +1,15 @@
+import gql from 'graphql-tag';
+
+export const GET_CELEBRATION_STEPS = gql`
```
OzzieOrca: Do we want these spread into separate files? I think it might be better to colocate them with the component that uses them (either in the same file or directory) and use fragments for subcomponents to specify what they need.

Webpack might have a loader for GraphQL files. Idk if it is needed.

src/dashboard/graphql/getMembers.js
```diff
@@ -0,0 +1,26 @@
...
+            members {
+                members_1W {
+                    data {
+                        stage
+                        members
+                        stepsAdded
+                        stepsCompleted
+                    }
+                }
+                members_default {
```
OzzieOrca: Camelcase? Most other JS stuff is and the GraphQL API is.

src/dashboard/insights.js
```diff
@@ -0,0 +1,70 @@
...
+    });
+
+    const cache = new InMemoryCache();
+    const stateLink = withClientState({
+        cache,
+        defaults,
+        resolvers,
+    });
+
+    const client = new ApolloClient({
```
OzzieOrca: This should go somewhere common eventually.

jest.config.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+    roots: ['src/dashboard'],
+    setupFiles: ['./src/dashboard/setupTests.js'],
```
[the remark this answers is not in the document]
bwtoptal: So you're thinking about changing this config to run tests in all directories not only `dashboard` directory with React components? I was thinking that we could keep all the React stuff in one folder, but it shouldn't be named `dashboard` of course and actually, it could be higher in a hierarchy.  Hmm... maybe we should create a separate directory in the top level and call it 'react-src' or sth like this, what do you think?

package.json
bwtoptal: Good point :) 👍

jest.config.js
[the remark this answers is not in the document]
OzzieOrca: I think if you could make that directory just be `./src` that would be great. There's a tiny bit of Angular in there but I don't see any tests so hopefully there aren't any issues just running the whole directory.

package.json
```diff
@@ -42,6 +54,12 @@
         "ng-file-upload": "^12.2.13",
         "ng-infinite-scroll": "~1.3.0",
         "papaparse": "^4.5.0",
+        "prop-types": "^15.7.2",
```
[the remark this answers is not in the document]
OzzieOrca: Let's get through what you've currently done work on and then add Typescript if we can. I can work on adding that tooling if you want. So don't worry about it for this PR.

package.json
```diff
@@ -73,11 +91,16 @@
         "@babel/runtime": "^7.4.3",
         "@brainly/onesky-utils": "^1.3.1",
         "angular-mocks": "^1.7.8",
+        "apollo-link-mock": "^1.0.1",
         "babel-eslint": "^10.0.1",
         "babel-loader": "^8.0.4",
         "babel-plugin-angularjs-annotate": "^0.10.0",
+        "babel-polyfill": "^6.26.0",
```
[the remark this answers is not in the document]
bwtoptal: Yeah, there was an error running tests with Enzyme, but since we're getting rid of Enzyme it can be removed too.

package.json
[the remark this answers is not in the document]
bwtoptal: Thanks, I'll be happy to add Typescript in a separate task. It's already quite a lot setup going on in this PR :)

package.json
bwtoptal: Yeah! That looks much better! Very impressive work you did there! 👍 I'll use a similar approach.

package.json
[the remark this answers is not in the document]
bwtoptal: Actually this error was thrown by running `jest`, I fixed that adding:
```json
{
                "targets": {
                    "node": "current"
                }
            }
```
in `.babelrc`

src/dashboard/components/navigation.spec.js
bwtoptal: Great idea! 👍

src/dashboard/graphql/getCelebrations.js
bwtoptal: I agree. It's much cleaner to keep them together with components. I remove the ones I don't use at the moment. 👍

package.json
bwtoptal: Happy to try a different tool 👍

package.json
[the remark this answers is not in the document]
bwtoptal: eh.. Unfortunately adding this, breaks Angular.js runtime app... So I guess I'll stick with `babel-polyfill` for now. Basically `jest` needs `regeneratorRuntime` which is included in `babel-polyfill`.

jest.config.js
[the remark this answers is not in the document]
bwtoptal: Also, I'll modify a little bit naming and structure conventions when pushing a PR with some more components. Do you think it would be ok, to move all the Angular files from `./src` directory and keep only React stuff there?

.travis.yml
```diff
@@ -6,6 +6,9 @@ addons:
 
 script:
     - yarn test
+    - yarn global add apollo
```
OzzieOrca: This is so you can run `gql:schema`? Just add `apollo` as a dev dependency.

src/dashboard/apolloClient.js
```diff
@@ -0,0 +1,29 @@
...
+    defaults,
+    resolvers,
+});
+
+export const createApolloClient = token => {
+    const httplink = createHttpLink({
+        uri: 'https://api-stage.missionhub.com/apis/graphql',
+        includeExtensions: true,
+        headers: {
+            Authorization: `Bearer ${token}`,
```
OzzieOrca: The server will update this token over time. https://www.apollographql.com/docs/react/recipes/authentication/#header seems to be a way to run something before every request. Idk if going all the way into localStorage every time is best. It's probably fast enough. I think Angular keeps it in memory. Idk if there's a way to get at that. Hybrid is fun :)

Eventually we'll need to pull the new token out of the `x-mh-session` header of every response and update it if it's changed. I haven't done that to mobile yet...

src/dashboard/testUtils/index.js
```diff
@@ -0,0 +1,25 @@
...
+    // All the props need to stay the same otherwise render won't work.
+    const wrapper = ({ children }) => (
+        <ApolloProvider client={mockApolloClient}>{children}</ApolloProvider>
+    );
+
+    const renderResult = render(component, { wrapper });
+
+    return {
+        ...renderResult,
+        rerender: component => renderResult.update(component),
```
OzzieOrca: I don't think this line is needed for this library. `renderResult` contains `rerender`. I had some more custom code for when the `wrapper` option wasn't supported in `react-native-testing-library`. I think I can remove it from the mobile app too now.

src/dashboard/testUtils/index.js
```diff
@@ -0,0 +1,25 @@
...
+import { ApolloProvider } from 'react-apollo-hooks';
+import { render } from '@testing-library/react';
+
+import { createApolloMockClient } from './apolloMockClient';
+
+export const renderWithContext = (component, { mocks }) => {
+    const mockApolloClient = createApolloMockClient(mocks);
+
+    // Warning: don't call any functions in here that return new instances on every call.
+    // All the props need to stay the same otherwise render won't work.
```
OzzieOrca: Can you make this say:
```suggestion
    // All the props need to stay the same otherwise rerender won't work.
```
I wrote it wrong... I was having issues of the props on the provider changing and rerendering the entire component tree, not just the component under test.

src/setupTests.js
```diff
@@ -0,0 +1,5 @@
+import 'babel-polyfill';
```
OzzieOrca: Ah this is only used for tests? That's better.

Sorry to come back to this again...

Just curious, `import '@babel/runtime';` doesn't work?

`babel-polyfill` only goes to v6. v7 is at`@babel/polyfill` but apparently it's been deprecated https://babeljs.io/docs/en/babel-polyfill. Looks like it suggests doing
```
import "core-js/stable";
import "regenerator-runtime/runtime";
```
instead. Idk if both are needed to solve you issue. Node should be new enough to support generators. https://kangax.github.io/compat-table/es6/#test-generators

Idk enough here. If trying a couple more trivial things doesn't work, just go with it.

bwtoptal: Actually, I can't even import `import '@babel/runtime';`... Based on what I read usually we would just specify target env in `.babelrc` and node already has everything needed. However, if I do that, then AnguarJS app doesn't start and tbh I didn't investigate why exactly. 

Looks like adding only: 
```js
import "regenerator-runtime/runtime";
```
solves the issue 😄

src/dashboard/apolloClient.js
bwtoptal: I need to read more about Apollo local state. I saw that you use Redux in React Native app as a state container? 

I think it would be possible to simply take it from localStorage once, and then store it in the local state (Apollo or Redux). Then we wouldn't need to get it from localStorage on every request made. What do you think?

OzzieOrca: I haven't gotten time to become familiar with Apollo local state stuff yet. If it seems nice to use that might work.

I think Angular and React are going to have to share that state until one of them is dead. Both REST GraphQL responses can come back with an updated JWT in the `x-mh-session` header. Could we just use the Angular service for now? Is that easy to get to? We could add a setter method to it.

I'm hesitant to add in a state management library like Redux with only one screen... Maybe we decide we need it but it's a lot of boilerplate. If there's small stuff we can share it with `React.createContext` and the `useContext` hook. Might have to do that to share Angular services in deeply nested React components.

Idk if we're there yet since we only have one route and screen but seems like using `react2angular` creates almost the equivalent of a new React app. Might have to use context and a wrapping component on each to share things that would normally be configured at the root of the React component tree in `App.js`.

src/setupTests.js
OzzieOrca: Ok cool

jest.config.js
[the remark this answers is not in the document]
OzzieOrca: You could move them out... I don't think they'll mess anything up staying there. A while ago I was thinking about moving all the Angular stuff in there but I never did. I guess that might be a nice separation.

.travis.yml
OzzieOrca: Looks like travis is unhappy... I wonder if not specifying `graphql` as a direct dependency would work... Otherwise I guess we should specify how to resolve it or go back to what you had. Don't think I've run into that before...

bwtoptal: Yeah... I was looking at this for a while. Looks like adding `apollo` as a dependency causes to have an additional dependency for `graphql` package which causes the error... I was trying to lock the version etc but no luck... I have no idea why in React Native app it works. I tried using exact same versions... I'll keep looking.

jest.config.js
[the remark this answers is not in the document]
bwtoptal: I'll move everything Angular related to `./app` directory and leave only React stuff in `src`.

OzzieOrca:
Can you leave the `src/.well-known` and `src/i18n` directories where they were? `.well-known` is just static stuff needed for mobile app links managed by webpack that we want to stay around in the new code. All the i18next config should be sharable React. I just stole the config from our React Native codebase.

I think it would be better to move the other files in a new PR we can merge immediately. This PR or a branch containing it may be around for a little bit until we're ready to release the insight dashboard.

bwtoptal:
Sure, I can leave `src/.well-known` and `src/i18n` where they currently are, but do you want me to create a separate PR with just moving these files out of `src` directory so we could merge them? 

So we want to have this branch as the main branch for all the Dashboard related tasks? Wouldn't be easier just to merge everything to master, but do not display a link to redirect to a new dashboard. This way we won't have a huge merge to do at the end :) What do you think?

src/dashboard/apolloClient.js
bwtoptal: Looks like we can simply get `x-mh-session` header from GraphQL responses and use Angular `authenticationService` to store a JWT token in a local storage. We already pass this as a dependency from Angular to React. 

I agree that adding Redux at this point might be too much. I'll check how the Apollo local state works or use context for sharing in React components.

.travis.yml
bwtoptal: Would it be ok if I leave this global install of apollo for now? I tried to get this working, but I'm running out of the ideas.. and I don't want to waste to much time on that.  I could come back to this later if that's fine.

package.json
```diff
@@ -30,9 +39,17 @@
...
+        "apollo-cache-inmemory": "^1.6.0",
+        "apollo-client": "^2.6.0",
+        "apollo-link": "^1.2.11",
+        "apollo-link-http": "^1.5.14",
+        "apollo-link-state": "^0.4.2",
         "bootstrap": "4.3.1",
         "countrily-data": "^1.0.1",
         "country-region-data": "^1.4.4",
+        "emotion-theming": "^10.0.10",
+        "graphql": "^14.3.1",
```
OzzieOrca: If you toss this, the `apollo` cli seems to work. Everything else seemed to work. So if anything else needs it I guess it imports it as its own dependency.

bwtoptal: Unfortunately when I remove this dependency, during startup `apollo-link-state` throws an error about missing dependency to `graphql`... so I guess It should stay..

OzzieOrca: That's weird... I'm on yarn `1.16.0` and it works.
Screen Shot 2019-06-14 at 10 32 52 AM

But one global install's not going to hurt anything.

src/setupTests.js
```diff
@@ -0,0 +1,5 @@
+import 'regenerator-runtime/runtime';
+
+jest.mock('./dashboard/apolloClient', () => ({
+    apolloClient: require('./dashboard/testUtils/apolloMockClient').createApolloMockClient(),
+}));
```
OzzieOrca: https://github.com/CruGlobal/missionhub-react-native/blob/4284ff60fa0df28c5d4b93e6f1e9e1f92489f51d/__mock__/resetGlobalMockSeeds.ts#L1 also needs to be run here otherwise test data can change depending on the number of tests running.

src/setupTests.js
```diff
@@ -1,5 +1,8 @@
 import 'regenerator-runtime/runtime';
+import { resetGlobalMockSeeds } from './dashboard/testUtils/globalMocks';
 
 jest.mock('./dashboard/apolloClient', () => ({
     apolloClient: require('./dashboard/testUtils/apolloMockClient').createApolloMockClient(),
 }));
+
+resetGlobalMockSeeds();
```
OzzieOrca: It needs the `beforeEach` so the seeds get reset before every test.

bwtoptal: Cool! I thought it will be enough to run it before each test file but changed it to run before each test.

OzzieOrca: Ya if you add new tests or only run certain tests the mock data given to a specific test can change.

PR #196, opened 2019-05-31
MHP-2688: Remove person.user.primary_organization_id and auth_response.recent_organization

OzzieOrca:
These are leftover from legacy and they are only being used to log to Rollbar and to hide the subnav for no good reason.

### DIFF
--- a/app/assets/javascripts/angular/services/authentication.service.js
+++ b/app/assets/javascripts/angular/services/authentication.service.js
@@ -42,18 +42,14 @@ function authenticationService(
 
     const updateUserData = async () => {
         const me = await loggedInPerson.load();
-        const currentState = localStorageService.get('state');
-        const currentOrganization = currentState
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/angular/services/authentication.service.js
```diff
@@ -42,18 +42,14 @@ function authenticationService(
...
-        const currentState = localStorageService.get('state');
-        const currentOrganization = currentState
-            ? currentState.currentOrganization
-            : null;
 
-        setState(currentOrganization, me);
+        setState(me);
     };
 
     const setupUserSettings = async organization => {
```
bwtoptal: Parameter `organization` is not used anymore.

OzzieOrca: Good catch :)

PR #199, opened 2019-06-07
MHP-2651: Dashboard navigation

bwtoptal:
@OzzieOrca, I would be interested in your overall opinion. 

I still have a few questions/issues to address:
- [x] Do you think I should add Redux for keeping app state or go with Apollo local state: https://www.apollographql.com/docs/react/essentials/local-state/ ?
- [x] Should we implement a dropdown menu after clicking on the User Avatar? 
- [x] WIP: I'm having some issues navigating between Angular and React links...
- [x] WIP: I'm having some issues with running tests using Jest, probably missing some file loaders


OzzieOrca:
I didn't realize the designs were redoing the whole nav. I started a conversation with Alex about that.

I think we should redo the nav or get the new insights dashboard out there and not do both at the same time. Seems like design loves to redo the menu without telling me :)

If we go the nav route, I wonder how much work it would be to convert the parent app to a React app, use react-router and then use `react2angular` for all the components. I sorta assumed we'd save that for the end but might be easier to do it now instead of trying to run 2 navigators. I think ui-router does have some support for React though...

bwtoptal:
The way I do it now is that I added a new Route for `insights` which actually runs on a separate template: 
```js
.state({
                name: 'appReact.insights',
                url: '/ministries/:orgId/insights/personal',
                component: 'insights',
                resolve: {
                    orgId: $transition$ => $transition$.params().orgId,
                },
            })
```
There are few routes which we need in React so there is additional state which points to the same component for all subroutes of `/ministries/:orgId/insights/:subroutes`. Both states point to the same component which is `react2angular`  insight component (probably it would need a better name), so the rest of navigation is handled by React router. 

My idea was that this new navigation would be currently visible only on this one `insights` route, for the dashboard. After we're done with it we could move the next screens, one by one to the React app. And just redirect them in Angular routes config.

bwtoptal:
The data from GraphQL is mocked here, I still wait for https://jira.cru.org/browse/MHP-2618 to be deployed.

src/dashboard/containers/MemberStagesChart/index.js
```diff
@@ -0,0 +1,115 @@
...
+
+    if (loading) {
+        return <Wrapper>Loading...</Wrapper>;
+    }
+
+    const { stages_report: report } = data;
+
+    const MEMBERS_LABEL = 'Members';
+    const PERSONAL_STEPS_LABEL = 'Personal Steps of Faith';
+    const STAGE_LABEL = 'Stage';
```
OzzieOrca: These should be translated
```js
import i18next from 'i18next';

const STAGE_LABEL = i18next.t('namespace:stage');
```

Or you can do it in the React component with
```js
import { useTranslation } from 'react-i18next';

const MyComponent = () => {
 const { t } = useTranslation('namespace')

  return <span>{t('stage')</span>;
};
```

I think all you need to do to get `react-i18next` working is [this](https://github.com/CruGlobal/missionhub-react-native/blob/develop/src/i18n/index.js#L40) and [this](https://github.com/CruGlobal/missionhub-react-native/blob/develop/src/i18n/index.js#L59-L62).

The graph data is a little weird so do what you need to do.

src/dashboard/containers/Navigation/index.js
```diff
@@ -0,0 +1,59 @@
...
+        text-decoration: none;
+    }
+
+    &.active {
+        color: ${props => props.theme.colors.highlight};
+    }
+`;
+
+const Navigation = () => {
+    const { orgId } = React.useContext(AppContext);
```
OzzieOrca: You could destructure this in the import:
```js
import React, { useContext } from 'react';```

File: src/dashboard/containers/PersonalStepsPage/index.js [hunk 3]
```diff
@@ -0,0 +1,37 @@
+// Vendors
+import React from 'react';
+import styled from '@emotion/styled';
+// Project
+import Card from '../../components/Card';
+import MemberStagesChart from '../MemberStagesChart';
+
+const Wrapper = styled.div``;
```
OzzieOrca: Still needed/needed for the future? If not you could just use a `<div>` or `<>`.

src/dashboard/containers/PersonalStepsPage/index.js
```diff
@@ -0,0 +1,37 @@
...
+
+const Index = () => {
```
OzzieOrca: This isn't a very descriptive name.

src/dashboard/containers/StepsOfFaithPage/index.js
```diff
@@ -0,0 +1,19 @@
+// Vendors
+import React from 'react';
+import styled from '@emotion/styled';
+// Project
+import Card from '../../components/Card';
+
+const Wrapper = styled.div``;
```
OzzieOrca: Still needed?

src/dashboard/defaultTheme.js
```diff
@@ -0,0 +1,40 @@
...
+        size: '14px',
+    },
+    colors: {
+        primary: '#505256',
+        secondary: '#B4B6BA',
+        highlight: '#3CC8E6',
+        highlightDarker: '#007398',
+        white: '#ffffff',
+    },
+    graph: {
```
OzzieOrca: This going to be used a lot? Just wondering if it rises to the level of a "Theme". It might...

src/dashboard/testUtils/index.js
```diff
@@ -3,14 +3,21 @@ import { ApolloProvider } from 'react-apollo-hooks';
 import { render } from '@testing-library/react';
 
 import { createApolloMockClient } from './apolloMockClient';
+import defaultTheme from '../defaultTheme';
+import { ThemeProvider } from 'emotion-theming';
+import AppContext from '../appContext';
 
-export const renderWithContext = (component, { mocks }) => {
+export const renderWithContext = (component, { mocks } = {}, context = {}) => {
```
OzzieOrca: I think it would be nicer to keep the number of parameters down here since you may not use them all each time but would have to provide null or something if you wanted to skip one.
```suggestion
export const renderWithContext = (component, { mocks, context = {} } = {}) => {
```

Also `context` seems a little general. `appContext` maybe? Idk. Not sure how many values we plan on sticking in here...

app/assets/javascripts/angular/components/organizationOverview/organizationOverview.html
```diff
@@ -120,7 +120,7 @@
                         </li>
                         <!--<li>
                             <a
-                                ui-state="'appWithoutMenus.insights'"
+                                ui-state="'appReact.insights'"
```
OzzieOrca: This is the wrong state after you renamed it.
```suggestion
                                ui-state="'app.ministries.ministry.insights'"
```

app/assets/javascripts/angular/components/react/appReact.component.js
```diff
@@ -0,0 +1,8 @@
+import template from './appReact.html';
+
+angular.module('missionhubApp').component('appReact', {
```
OzzieOrca: Is this component needed? Doesn't seem like you're using it...

src/dashboard/containers/Layout/index.js
```diff
@@ -0,0 +1,56 @@
...
+    line-height: 20px;
+    flex-direction: column;
+    background-color: #f2f2f2;
+`;
+
+const Content = styled.div`
+    width: 100%;
+`;
+
+const Layout = () => {
```
OzzieOrca: This may be too general of a name for a nav that is only for insights. But I guess we can deal with that when we add more components.

src/dashboard/components/Card/__tests__/__snapshots__/Card.js.snap
```diff
@@ -0,0 +1,63 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`<Card /> should render properly 1`] = `
+<div>
+  <div
+    class="css-11s8wh0"
```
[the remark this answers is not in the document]
bwtoptal: Yes, it hashes the styles. If any styles are changed the hash is different and then we need to regenerate snapshots.

src/dashboard/defaultTheme.js
bwtoptal: To be honest, I'm not sure yet if it's going to be used a lot... When I add other graphs/charts I'll see which parts of styling are common and how to reorganize it. Not sure if it stays here. Moving this to a separate file also make sense.

package.json
```diff
@@ -16,6 +16,7 @@
         "test-watch": "karma start",
         "test-debug": "karma start --browsers Chrome",
         "lint": "eslint '@(app|src)/**/*.js'",
+        "lint:fix": "eslint './*.{js,ts,tsx}' './{src,app}/**/*.{js,ts,tsx}' --fix",
```
OzzieOrca: Let's make the file patterns for `lint` and `lint:fix` match. Your new one is good.

src/dashboard/appContext.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+const AppContext = React.createContext(null);
```
OzzieOrca: Would `{}` be a better default value?
```suggestion
const AppContext = React.createContext({});
```

src/dashboard/containers/MemberStagesChart/index.js
```diff
@@ -0,0 +1,114 @@
...
+
+const Wrapper = styled.div`
+    height: 400px;
+`;
+
+const MemberStagesChart = props => {
+    const { data, loading } = useQuery(GET_STAGE_REPORT);
+
+    if (loading) {
+        return <Wrapper>Loading...</Wrapper>;
```
OzzieOrca: i18n

src/dashboard/containers/Navigation/index.js
```diff
@@ -0,0 +1,58 @@
...
+        color: ${({ theme }) => theme.colors.highlight};
+    }
+`;
+
+const Navigation = () => {
+    const { orgId } = useContext(AppContext);
+    return (
+        <NavigationBar>
+            <Link to={`/ministries/${orgId}/insights/personal`}>
+                Personal Steps
```
OzzieOrca: i18n

src/dashboard/containers/Navigation/index.js
```diff
@@ -0,0 +1,58 @@
...
+            </Link>
+            <Link to={`/ministries/${orgId}/insights/steps`}>
+                Steps of Faith
```
OzzieOrca: i18n

src/dashboard/containers/Navigation/index.js
```diff
@@ -0,0 +1,58 @@
...
+            </Link>
+            <Link to={`/ministries/${orgId}/insights/interactions`}>
+                Interactions
```
OzzieOrca: i18n

src/dashboard/containers/Navigation/index.js
```diff
@@ -0,0 +1,58 @@
...
+            </Link>
+            <Link to={`/ministries/${orgId}/insights/challenges`}>
+                Challenges
```
OzzieOrca: i18n

src/dashboard/containers/StepsOfFaithPage/index.js
```diff
@@ -0,0 +1,16 @@
+// Vendors
+import React from 'react';
+
+import Card from '../../components/Card';
+
+const StepsOfFaithPage = () => {
+    return (
+        <div>
+            <Card title={'Steps of Faith'} subtitle={'To be implemented'} />
```
OzzieOrca: i18n

Is subtitle still going to be used?

src/dashboard/insights.js
```diff
@@ -4,34 +4,39 @@ import PropTypes from 'prop-types';
...
+import { createApolloClient } from './apolloClient';
+import AppContext from './appContext';
+import defaultTheme from './defaultTheme';
+import Layout from './containers/Layout';
 
 const Insights = ({ orgId, authenticationService }) => {
     const token = authenticationService.isTokenValid();
     const apolloClient = createApolloClient(token);
 
+    i18n.use(initReactI18next).init({
```
OzzieOrca: Unless it's needed here, can you stick this in https://github.com/CruGlobal/missionhub-web/blob/master/src/i18n/i18next.config.js.

Should be able to [chain it](https://github.com/CruGlobal/missionhub-react-native/blob/develop/src/i18n/index.js#L40) and use the existing init config block to add the [react config](https://github.com/CruGlobal/missionhub-react-native/blob/develop/src/i18n/index.js#L59-L62). This looks like it will initialize it twice which might have unexpected results.

src/setupTests.js
```diff
@@ -1,5 +1,16 @@
 import 'regenerator-runtime/runtime';
+import i18n from 'i18next';
+import { initReactI18next } from 'react-i18next';
+
 import { resetGlobalMockSeeds } from './dashboard/testUtils/globalMocks';
+import './i18n/i18next.config';
+
+i18n.use(initReactI18next).init({
```
OzzieOrca: After moving the react i18next setup, we can delete this.

src/dashboard/insights.js
bwtoptal: You are right... Thanks! 👍

src/dashboard/containers/MemberStagesChart/index.js
```diff
@@ -0,0 +1,116 @@
...
+    const { data, loading } = useQuery(GET_STAGE_REPORT);
+    const { t } = useTranslation('insights');
+
+    if (loading) {
+        return <Wrapper>{t('loading')}</Wrapper>;
+    }
+
+    const { stages_report: report } = data;
+
+    const MEMBERS_LABEL = i18n.t('insights:members');
```
OzzieOrca: Since these are in the render function, they could just use `t` from `useTranslation` and not have to have the namespace each time.

PR #200, opened 2019-06-24
MHP-2510: Unsubscribe landing page

bwtoptal:


app/assets/javascripts/angular/components/unsubscribe/unsubscribe.html
```diff
@@ -10,7 +10,7 @@ <h1>
             </h1>
             <h2>
                 {{ 'unsubscribe:noMoreEmails' | t }} {{ $ctrl.organizationName
-                }}
+                }}.
```
OzzieOrca: You could just stick all of this in the translation and use a variable for the org name.
https://www.i18next.com/translation-function/interpolation

Angular should pass options from the filter on to `i18next.t`.
https://github.com/CruGlobal/missionhub-web/blob/master/app/assets/javascripts/angular/filters/translation.filter.js#L5

PR #201, opened 2019-06-25
MHP-2456: Add handling for sign up screen with an invite link

bwtoptal:


app/components/authentication/signIn.html
```diff
@@ -1,27 +1,61 @@
...
+                href
+                class="btn-primary btn-facebook f5"
+                title="{{ 'login.useFacebook' | t }}"
+                ng-click="$ctrl.facebookService.signIn()"
+            >
+                {{ 'login.useFacebook' | t }}
+            </a>
+        </p>
+    </div>
+    <div ng-if="!$ctrl.inviteState">
```
OzzieOrca: We'll see what Matt and Alex say but hopefully we can reuse 2 of these buttons and not duplicate the HTML. Even as is, I think you could share the HTML but just change the label conditionally.

bwtoptal:
I pushed texts changes based on the discussion in Jira comment: https://jira.cru.org/browse/MHP-2456?focusedCommentId=81608&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-81608 
and also applied suggested fixes.

src/i18n/locales/en-US.js
```diff
@@ -417,7 +417,11 @@ export default {
             login: 'Log in',
             login_with_key: 'Log in with The Key',
             login_with_facebook: 'Log in with Facebook',
-            invite_in_progress: 'Please login to finish your invite.',
+            invite_in_progress:
+                'To complete your invite, create account or log in.',
+            signUpWithEmail: 'Sign Up with Email',
+            signInWithEmail: 'Sing In with Email',
```
OzzieOrca:
```suggestion
            signInWithEmail: 'Sign In with Email',
```

app/components/authentication/signIn.html
```diff
@@ -1,27 +1,41 @@
...
             href
             class="btn-key btn-primary f5"
-            title="{{ 'login.login_with_key' | t }}"
+            title="{{ $ctrl.inviteState ? 'login.signInWithEmail' : 'login.login_with_key' | t }}"
             target="_self"
             ng-href="{{$ctrl.theKeyUrl}}"
         >
-            {{ 'login.login_with_key' | t }}
+            {{ $ctrl.inviteState ? 'login.signInWithEmail' :
+            'login.login_with_key' | t }}
```
OzzieOrca: I think all these can just be the new values. You can run it by product if you're unsure.

bwtoptal: Good point actually! 👍 I asked the question in a comment.

PR #202, opened 2019-06-27
MHP-2652: Updated designs for Members & Personal stages graphs

bwtoptal:


app/assets/javascripts/angular/components/organizationOverview/organizationOverview.scss
```diff
@@ -63,6 +63,9 @@ organization-overview {
             display: block;
         }
     }
+
+    .insights-tab {
+    }
```
OzzieOrca: Is something supposed to be in here?

src/dashboard/components/BarChart/__tests__/Card.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+import { renderWithContext } from '../../../testUtils';
+import Card from '../';
+
+describe('<Card />', () => {
```
OzzieOrca: Shouldn't this be testing BarChart? That's the folder this test is in...

src/dashboard/components/Header/index.js
```diff
@@ -0,0 +1,25 @@
...
+    color: ${({ theme }) => theme.colors.highlightDarker};
+    margin-bottom: 0px;
+    margin-top: 48px;
+    padding-left: 21px;
+`;
+
+const Header = props => {
+    const { title } = props;
+
+    return <Title>{title}</Title>;
```
OzzieOrca: Seems like just returning `Title` directly might be more flexible. Then you can pass in whatever children you want.

bwtoptal:
Yes, I think Alex could test the overall layout, styles etc. It's still not connected to the backend, because it waits for https://jira.cru.org/browse/MHP-2618 to be done.

src/dashboard/containers/PersonalStepsPage/index.js
```diff
@@ -11,11 +11,9 @@ const PersonalStepsPage = () => {
 
     return (
         <div>
-            <Header
-                title={
-                    'Together we have taken 1,234 personal steps of faith in 2019.'
-                }
-            />
+            <Header>
+                Together we have taken 1,234 personal steps of faith in 2019.
```
OzzieOrca: I'm guessing this will change once it has data but it needs i18n.

src/dashboard/components/BarChart/__tests__/__snapshots__/BartChart.js.snap
```diff
@@ -0,0 +1,17 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`<BartChart /> should render properly 1`] = `
+<div>
+  <div
+    style="width: 100%; height: 100%;"
+  />
```
OzzieOrca: This test just seems to render an empty div which doesn't seem too helpful. Idk how to make the chart actually render something...

src/dashboard/containers/PersonalStepsPage/index.js
bwtoptal: Exactly, I left it because it needs data. I'll fix that when connecting to the data source.

src/dashboard/components/BarChart/__tests__/__snapshots__/BartChart.js.snap
bwtoptal: Actually, it's a good point. In runtime, I can see that the graph is simply a SVG element, so I guess there should be a way to generate it in tests as well. I'll play around to see if it's possible and push the changes together with connecting it to the data source.

OzzieOrca: Sounds good. I'm not sure if the issues if Jest not rendering the SVG or that the graph has no data.

src/dashboard/containers/PersonalStepsAddedChart/__tests__/PersonalStepsAddedChart.js
```diff
@@ -10,38 +10,30 @@ describe('<PersonalStepsAddedChart />', () => {
...
-                                    __typename: 'Data',
-                                    pathway_stage: 'NOT SURE',
-                                    member_count: 11,
-                                    steps_added_count: 34,
-                                    steps_completed_count: 37,
-                                }
-                            ])
-                        })
-                    })
+                        OrganizationStageReport: () =>  ([
```
bwtoptal: I was trying for a while to mock response from Apollo Client but failed. I looked at your code in React Native app and it looks similar. Maybe you know what may be wrong here? I'd really appreciate that.

OzzieOrca: If you go through `Query`, you have to use the relationship names instead of the type names. So
```diff
- OrganizationStageReport: () =>  ([
+ organizationPathwayStagesReport: () =>  ([
```

I wish it would throw some more errors when you try to mock something that doesn't exist...

If you want a bunch of the same thing you can use https://www.apollographql.com/docs/graphql-tools/mocking/#using-mocklist-in-resolvers. 

You could also mock the type instead of the `Query` type if you wanted them to all be the same or be a random number:
```ts
mocks: {
  OrganizationStageReport: () => ({ memberCount: 1 }) // or some faker random value that would be different for each one...
}
```

PR #203, opened 2019-07-03
MHP-2653: Personal Steps Completed 

bwtoptal:


bwtoptal:
Changes from MHP-2652 are also here because I need them to not have too many conflicts, so only da7511d commit would need a code review. This chart is also not connected to the GraphQL endpoint yet. I need to talk with @ckozus about it.

src/dashboard/components/BarChart/index.js
```diff
@@ -0,0 +1,83 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { ResponsiveBar } from '@nivo/bar';
+import { withTheme } from 'emotion-theming';
+import { line } from 'd3-shape';
+
+const BarChart = props => {
+    const { theme, data, keys, indexBy, average, tooltip } = props;
+
+    const AverageLine = ({ bars, xScale, yScale }) => {
```
OzzieOrca: What's this the average of? The number of steps in a day?

https://nivo.rocks/storybook/?path=/story/bar--with-marker looks like it supports markers but I guess that doesn't help you calculate anything.

My initial thought is this is a bit of code for drawing a line. But if that's how Nivo wants you to do it, that's fine.

src/dashboard/components/SwitchButton/index.js
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    &.active {
+        background-color: ${({ theme }) => theme.colors.highlight};
+        color: ${({ theme }) => theme.colors.white};
+    }
+`;
+
+const SwitchButton = ({ leftLabel, rightLabel, onLeftClick, onRightClick }) => {
+    const [isLeftActive, setLeftActive] = React.useState(true);
```
OzzieOrca: I kinda like importing these specifically... Idk. Doesn't matter.

src/dashboard/containers/PersonalStepsCompletedChart/index.js
```diff
@@ -0,0 +1,227 @@
...
+                    average={average}
+                    tooltip={tooltip}
+                />
+                <PreviousButton onClick={onPreviousRangeClick} />
+                <NextButton onClick={onNextRangeClick} />
+            </Wrapper>
+            <Footer>
+                <SwitchButton
+                    leftLabel={'1M'}
+                    rightLabel={'1Y'}
```
OzzieOrca: These might need to be translated... Idk. They're a little weird...

src/dashboard/components/BarChart/index.js
bwtoptal: Unfortunately... That's how they proposed in their GitHub issues: https://github.com/plouc/nivo/issues/139

I would like to have an easier way of doing that.

src/dashboard/containers/PersonalStepsReachedInfo/__tests__/PersonalStepsReachedInfo.js
```diff
@@ -0,0 +1,32 @@
...
+            {
+                mocks: {
+                    Query: () => ({
+                        impactReport: () => ({
+                            pathwayMovedCount: () => 13,
+                        }),
+                    }),
+                },
+                appContext: {
+                    orgId: 1,
```
OzzieOrca: @bwtoptal This was the issue. Something seems to be swallowing errors but the `organizationId` variable being passed into the query was undefined which I guess results in a failed query. The error key contains it. I found it by making the debugger "Pause on exceptions".
Screen Shot 2019-07-19 at 1 10 13 PM

package.json
```diff
@@ -49,9 +49,10 @@
         "bootstrap": "4.3.1",
         "countrily-data": "^1.0.1",
         "country-region-data": "^1.4.4",
+        "d3-shape": "^1.3.5",
         "emotion-theming": "^10.0.10",
         "eslint-plugin-import": "^2.17.3",
-        "graphql": "^14.3.1",
+        "graphql": "^14.4.2",
```
OzzieOrca: This seemed to fix the errors you were getting adding apollo.

`yarn gql:schema` needs to be run for the default mocks to work I believe. I don't know if you had that schema.json downloaded when you were testing. Actually I guess it should throw an error on import so nvm but good to keep it up to date...

OzzieOrca:
Seems like you have a failing test. The i18next config can't find a file.

src/dashboard/containers/PersonalStepsReachedInfo/__tests__/PersonalStepsReachedInfo.js
bwtoptal: Wow! That's awesome! Thanks a lot @OzzieOrca 👍

package.json
bwtoptal: Yes, looks like it solves the issue... Yeah, I had the schema.json downloaded during testing.

bwtoptal:
I don't think that's the issue, i18n, in fact, throws a warning but it works...  The issue was that for some reason when both tests were run one after each other the second one failed... If I ran them separately both work. I resolved the issue by passing `orgId` in the context for the first test... As you noticed before, for some reason exceptions are weirdly handled...

src/dashboard/containers/PersonalStepsCompletedChart/__tests__/PersonalStepsCompletedChart.js
```diff
@@ -0,0 +1,14 @@
+// Vendors
```
OzzieOrca: Can you remove this?

src/dashboard/containers/PersonalStepsTakenInfo/__tests__/PersonalStepsTakenInfo.js
```diff
@@ -0,0 +1,38 @@
+// Vendors
```
OzzieOrca: Remove

bwtoptal:
Changes merge to master in PR: https://github.com/CruGlobal/missionhub-web/pull/256

PR #205, opened 2019-07-18
MHP-2804: Admin redirect

OzzieOrca:


OzzieOrca:
@sheldond This doesn't seem to be working. After getting a 201 from `https://api-stage.missionhub.com/admin/auth` and redirecting to `https://api-stage.missionhub.com/admin`, I get a 302 back to `http://stage.mhub.cc/` (see https://github.com/CruGlobal/missionhub-api/pull/946#discussion_r304706529). Is the GET to `/admin` supposed to pass anything?

You can test it at https://deploy-preview-205--missionhub.netlify.com/admin. Just be aware it redirects back to `stage.missionhub.com` so you need to manually navigate back to netlify again.

Also any reason the first request isn't under the `api/v4` namespace?

sheldond:
Thanks for helping with this! The 302 back means that the cookie auth didn't work, so API is redirecting back to the login on web as there is no longer any way to login directly on the API domain.

After the POST responds 201 it sends a cookie session back, I can see the cookie in the `set-cookie` response header. But when the next GET request goes out it doesn't seem to include the cookie, therefore the next response is a 302 back to the login. Does web need to do something special in that request to include the cookie?

The reason this endpoint is not namespaced under api/v4 is because the rest of the admin is also outside the api/v4 namespace and I wanted consistency by grouping all admin related things under the /admin namespace. We could move it if you feel like that makes more sense to you!

PR #206, opened 2019-07-18
Mhp 2762 add confirmation modals

TheNoodleMoose:


app/assets/javascripts/angular/components/organizationCleanup/organizationCleanup.component.js
```diff
@@ -21,10 +22,73 @@ function organizationCleanupController(
...
+                try {
+                    const DateResponse = await DateModal.result;
+                    archiveBefore(archiveDate);
+                } catch (error) {
+                    return null;
+                }
+                break;
+            }
+            default:
+                return null;
```
OzzieOrca: You could probably just return undefined here and above unless something is expecting null
```suggestion
                return;
```

app/assets/javascripts/angular/components/organizationCleanup/organizationCleanup.component.js
```diff
@@ -21,10 +22,73 @@ function organizationCleanupController(
     this.archiveBeforeDate = new Date();
     this.archiveInactivityDate = new Date();
     this.checkIcon = checkIcon;
+    this.warningIcon = warningIcon;
+
+    // Opens up modal depending on the type
+    const confirmModal = async (type, archiveDate) => {
```
OzzieOrca: This is a lot of code for a confirm modal and there's a lot of similar code. Maybe making a `showConfirmModal(title: string): Promise` function that you could call before the archive would be a better solution. Idk, let me know what you think. You could also try to compute some stuff up top like the title and the function to call and then just call everything once.

app/assets/javascripts/angular/components/organizationCleanup/organizationCleanup.component.js
```diff
@@ -21,10 +22,73 @@ function organizationCleanupController(
...
+                        title: () =>
+                            $filter('t')(
+                                `Are you sure you want to archive these contacts by inactivity? `,
+                            ),
+                        closeLabel: () => $filter('t')('Ok'),
+                        dismissLabel: () => $filter('t')('Cancel'),
+                    },
+                    windowClass: 'pivot_theme',
+                    backdrop: 'static',
+                    keyboard: false,
```
OzzieOrca: Any reason for this? I think `true` would allow users to get out by pushing <kbd>ESC</kbd>.

app/assets/javascripts/angular/components/organizationCleanup/organizationCleanup.component.js
```diff
@@ -21,10 +22,73 @@ function organizationCleanupController(
...
+                let archiveInactive = date => {
+                    archiveContacts('leaders_last_sign_in_at')(date);
+                };
+
+                const InactiveModal = $uibModal.open({
+                    component: 'iconModal',
+                    resolve: {
+                        title: () =>
+                            $filter('t')(
+                                `Are you sure you want to archive these contacts by inactivity? `,
```
OzzieOrca: i18next is using key based translations. You have to go add a key to [en-US.js](https://github.com/CruGlobal/missionhub-web/blob/master/src/i18n/locales/en-US.js#L496) and then use that key here. https://www.i18next.com/translation-function/essentials

Also you can do `i18next.t('namespace:key.subKey')` directly if you `import i18next from 'i18next';`. The filter is just a [layer of indirection](https://github.com/CruGlobal/missionhub-web/blob/master/app/assets/javascripts/angular/filters/translation.filter.js#L5).

TheNoodleMoose: Ahh, this makes since. I was trying to figure out where this was coming from. I will change it to be inside of en-US.js

app/assets/javascripts/angular/components/organizationCleanup/organizationCleanup.component.js
TheNoodleMoose: I had grabbed this modal from the one below that was already present in the file. I will change it to allow keyboard because that makes sense to allow the user to do.

app/assets/javascripts/angular/components/organizationCleanup/organizationCleanup.component.js
TheNoodleMoose: For some reason if I have just return, prettier is getting rid of it? Is just having return undefined; okay or should I look into trying to turn off this prettier formatting option?

app/assets/javascripts/angular/components/organizationCleanup/organizationCleanup.component.js
TheNoodleMoose: Yeah, you are right. This code is repetitive, I already see a better way to do it. I'll cleanup it up :+1:

src/i18n/locales/en-US.js
```diff
@@ -502,10 +502,10 @@ export default {
                     'Go to the Contacts screen, select the people who have graduated from your ministry and label them a "Alumni". If your ministry hasn\'t created that label yet, you can do so by adding it in Manage Labels under Tools. ',
                 archive_by_date: 'Archive by date',
                 archive_by_date_description:
-                    'Archive contacts who have been added before:',
+                    'Are you sure you want to archive these contacts by date?',
                 archive_by_inactivity: 'Archive by inactivity',
                 archive_by_inactivity_description:
-                    'Archive people who have been inactive since:',
+                    'Are you sure you want to archive these contacts by inactivity?',
```
OzzieOrca: These confirm questions need new keys. This changed the descriptions on the main page too:
Screen Shot 2019-07-19 at 10 06 48 AM

app/assets/javascripts/angular/components/organizationCleanup/organizationCleanup.component.js
```diff
@@ -22,9 +22,38 @@ function organizationCleanupController(
...
+                : 'persons_inactive_since';
+        const archiveDate =
+            title === 'ministries.cleanup.archive_by_inactivity_description'
+                ? this.archiveInactivityDate
+                : this.archiveBeforeDate;
+
+        const confirmModal = $uibModal.open({
+            component: 'iconModal',
+            resolve: {
+                title: () => $filter('t')(`${title}`),
```
OzzieOrca: Title should already be a string...
```suggestion
                title: () => $filter('t')(title),
```

app/assets/javascripts/angular/components/organizationCleanup/organizationCleanup.component.js
```diff
@@ -22,9 +22,38 @@ function organizationCleanupController(
...
+                closeLabel: () => $filter('t')('general.ok'),
+                dismissLabel: () => $filter('t')('general.cancel'),
+            },
+            windowClass: 'pivot_theme',
+            backdrop: 'static',
+            keyboard: true,
```
OzzieOrca: `true` should be the default. You can remove this line.
```suggestion

```

app/assets/javascripts/angular/components/organizationCleanup/organizationCleanup.component.js
```diff
@@ -22,9 +22,38 @@ function organizationCleanupController(
...
+        });
+
+        try {
+            const modalResponse = await confirmModal.result;
+            archiveContacts(archiveValue)(archiveDate);
+        } catch (err) {
+            return null;
```
OzzieOrca: `return null` is similar to `return undefined` which is the same as `return` which is the same as not returning anything. I think you can just remove the line. Ya the catch block will be empty but that should be fine...

src/i18n/locales/en-US.js
TheNoodleMoose: Okay I see now, sorry about that... will add the new keys :+1:

src/i18n/locales/en-US.js
```diff
@@ -503,7 +503,11 @@ export default {
                 archive_by_date: 'Archive by date',
                 archive_by_date_description:
                     'Archive contacts who have been added before:',
+                archive_by_date_title:
```
OzzieOrca: I'm being nitpicky now but `archive_by_date_confirm_message` might be a better key name. Just cuz there are other similar things in here.

TheNoodleMoose: Agreed, I sometimes struggle to figure out the best naming conventions. Thanks for the suggestions :)

PR #207, opened 2019-07-19
MHP-2656: Steps of Faith page

bwtoptal:


bwtoptal:
Similarly to https://github.com/CruGlobal/missionhub-web/pull/203 it misses:
- [ ] Connection for GraphQL endpoint
- [ ] More unit tests

src/dashboard/components/StepsChart/__tests__/StepsChart.js
```diff
@@ -0,0 +1,96 @@
+import React from 'react';
+import gql from 'graphql-tag';
+import moment from 'moment';
+import { waitForElement } from '@testing-library/react';
+
+import { renderWithContext } from '../../../testUtils';
+import StepsChart from '../';
+
+const QUERY = gql`
```
OzzieOrca: Is this the same query as one in the source code? Could you export it from there and use it in the test? Copying it in a test is a good way for it to get out of date. I didn't go digging around too much to check if it's the same. I assume this might apply to other tests.

bwtoptal: It's similar to the one in source code, however for these components it doesn't matter what query we use it. I mean the query is independent of the component because we also provide a `mapData` function which maps results for this specific query. So even if we change the query in the source code this test will still work. Rest of the tests are similar, I tried to extract query dependency, up to the main components (page components) and these components are only responsible for calling the provided query, showing loading status, parsing the response using provided function and displaying the data.

OzzieOrca: Do you have tests that run all the queries? I can dig in a little more, it's just a big PR to look through.

I think you should at least just pick one of the existing queries to run the test in the sub component with if you aren't testing them all. A new query just for the test seems a little odd.

bwtoptal:
Changes merged to master in PR: https://github.com/CruGlobal/missionhub-web/pull/256

PR #244, opened 2019-07-25
MHP-2764-Add-Labels-To-Auto-Assign

TheNoodleMoose:
This also includes MHP-2763 where I changed to text description for each tab.


app/assets/javascripts/angular/components/assignedAltSelect/assignedAltSelect.service.spec.js
```diff
@@ -0,0 +1,32 @@
...
+
+    beforeEach(inject(function(_assignedAltSelectService_) {
+        assignedAltSelectService = _assignedAltSelectService_;
+
+        this.person = { id: 123 };
+    }));
+
+    describe('isMe', function() {
+        it('should check whether the person matches the logged in person', function() {
+            expect(assignedAltSelectService.isMe(this.person)).toBe(true);
```
OzzieOrca: This test is failing https://travis-ci.org/CruGlobal/missionhub-web/builds/563652300#L1286.

app/assets/javascripts/angular/components/assignedAltSelect/assignedAltSelect.component.js
```diff
@@ -0,0 +1,57 @@
+import template from './assignedAltSelect.html';
+import './assignedAltSelect.scss';
+
+angular.module('missionhubApp').component('assignedAltSelect', {
```
OzzieOrca: `Alt` meaning alternate? Is this only for labels? Maybe `assignedLabelsSelect`?

app/assets/javascripts/angular/components/assignedAltSelect/assignedAltSelect.component.js
```diff
@@ -0,0 +1,57 @@
...
+        organizationId: '<',
+        disabled: '<',
+        actionAfterSelect: '&',
+    },
+    controller: assignedAltSelectController,
+    template: template,
+});
+
+function assignedAltSelectController($scope, assignedAltSelectService) {
+    var vm = this;
```
OzzieOrca: If you use arrow functions everywhere, you don't have to do this `this` hack any more. Arrow functions don't create their own `this` unlike normal functions which cause collisions.

So below you could do `this.$onInit = () => {` and then use `this` instead of `vm` inside.

app/assets/javascripts/angular/components/assignedAltSelect/assignedAltSelect.html
```diff
@@ -0,0 +1,20 @@
...
+    <ui-select-match
+        placeholder="{{ 'application.assign_search.assign_label' | t }}"
+    >
+        <span>
+            {{ $item | personName }}
+        </span>
+    </ui-select-match>
+    <ui-select-choices repeat="person in $ctrl.labels track by person.id">
+        <span>
+            {{ person | personName }}
```
OzzieOrca: Ah found what's creating `person.name`. If this is only handling labels, you could just render label.name directly and not use the filter.

app/assets/javascripts/angular/components/assignedAltSelect/assignedAltSelect.html
```diff
@@ -0,0 +1,20 @@
...
+    ng-disabled="$ctrl.disabled"
+>
+    <ui-select-match
+        placeholder="{{ 'application.assign_search.assign_label' | t }}"
+    >
+        <span>
+            {{ $item | personName }}
+        </span>
+    </ui-select-match>
+    <ui-select-choices repeat="person in $ctrl.labels track by person.id">
```
OzzieOrca:
```suggestion
    <ui-select-choices repeat="label in $ctrl.labels track by label.id">
```

app/assets/javascripts/angular/components/surveyOverview/surveyOverview.scss
```diff
@@ -326,12 +326,20 @@ organization-contact-import {
...
+        }
+
         .tab {
             border-bottom: 1px solid transparent;
 
-            &.active {
-                //border-bottom: 1px solid $blue-1;
-            }
+            // &.active {
+            //border-bottom: 1px solid $blue-1;
```
OzzieOrca: If this isn't being used, just delete it.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -72,15 +72,55 @@ function surveyOverviewQuestionsController(
...
+                {
+                    include: 'labels',
+                },
+                {
+                    errorMessage: 'error.messages.surveys.loadQuestions',
+                },
+            )
+            .then(res => res.data.labels);
+
+        let currentLabels = [];
```
```suggestion
        const currentLabels = Labels.filter(label => labelIds.includes(label.id));
```

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -72,15 +72,55 @@ function surveyOverviewQuestionsController(
...
+                            ? r.label_ids.split(',')
+                            : [];
+                        return [...a2, ...labelIds];
+                    }, []);
+                    return [...a1, ...ids];
+                }, []),
+            ),
+        ];
+
+        const Labels = await httpProxy
```
OzzieOrca: Usually data is camelCase in JS.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -206,14 +249,55 @@ function surveyOverviewQuestionsController(
         loadSurveyData();
     };
 
+    this.addLabelToRule = async (question, rule) => {
+        const index = question.question_rules.indexOf(rule);
+
+        if (!question.question_rules[index].assign_to) return;
```
```suggestion
        if (!question.question_rules[index]. assigned_labels) return;
```
?

TheNoodleMoose: Yep, didn't catch that will change.

app/assets/javascripts/angular/components/surveyOverview/surveyOverview.scss
TheNoodleMoose: Will do

app/assets/javascripts/angular/components/assignedAltSelect/assignedAltSelect.component.js
TheNoodleMoose: Yeah this is a better name, should I changed assignedSelect to assignedPeopleSelect or leave it alone?

app/assets/javascripts/angular/components/assignedAltSelect/assignedAltSelect.service.spec.js
TheNoodleMoose: Ah, yeah I deleted the isMe function because i'm no longer checking for people. I may just delete the whole test.

app/assets/javascripts/angular/components/assignedAltSelect/assignedAltSelect.component.js
OzzieOrca: That might be nice.

app/assets/javascripts/angular/components/assignedAltSelect/assignedAltSelect.component.js
TheNoodleMoose: :+1:

app/assets/javascripts/angular/components/assignedLabelSelect/assignedLabelSelect.component.js
```diff
@@ -0,0 +1,53 @@
...
+                // Ignore empty searches
+                this.labels = [];
+                return;
+            }
+
+            assignedLabelSelectService
+                .searchLabels(search, this.organizationId)
+                .then(labels => {
+                    if (search === undefined) {
+                        labels.map(label => {
```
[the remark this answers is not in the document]
TheNoodleMoose: Yeah I think I had this when I was also getting people and was trying to keep all the people as well at the labels. I'll change this.

app/assets/javascripts/angular/components/assignedLabelSelect/assignedLabelSelect.service.js
```diff
@@ -0,0 +1,24 @@
+angular
+    .module('missionhubApp')
+    .factory('assignedLabelSelectService', assignedLabelSelectService);
+
+function assignedLabelSelectService(httpProxy) {
+    return {
+        // Get all the labels from the current organizations
+        searchLabels: (query, organizationId) => {
```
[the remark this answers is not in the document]
OzzieOrca: Still unused...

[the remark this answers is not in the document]
TheNoodleMoose: Yeah, I'll just remove it

app/assets/javascripts/angular/components/assignedLabelSelect/assignedLabelSelect.component.js
```diff
@@ -0,0 +1,53 @@
...
+                            this.labels = [...this.labels, label];
+                        });
+                    } else if (search !== undefined) {
+                        labels.map(label => {
```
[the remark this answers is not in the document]
TheNoodleMoose: I think i fixed the max call stack error when searching for people who were already added. I'll push it up and let me know what you think.

TheNoodleMoose:
I think i fixed the max call stack error when searching for people who were already added. I'll push it up and let me know what you think.

app/assets/javascripts/angular/components/assignedPeopleSelect/assignedPeopleSelect.service.js
```diff
@@ -10,12 +10,11 @@ function assignedPeopleSelectService(
 ) {
     return {
         // Search for people in a particular organization that match the query
-        searchPeople: function(query, organizationId, deduper) {
+        searchPeople: function(organizationId, deduper) {
             return httpProxy
                 .get(
                     modelsService.getModelMetadata('person').url.search,
                     {
-                        q: query,
```
OzzieOrca: The person query is paginated and will return I think at most 25 results. So we need to search using the API.

app/assets/javascripts/angular/components/assignedPeopleSelect/assignedPeopleSelect.component.js
```diff
@@ -19,22 +19,33 @@ function assignedPeopleSelectController(
...
                 return;
             }
-
-            assignedPeopleSelectService
-                .searchPeople(search, this.organizationId, requestDeduper)
-                .then(people => {
-                    this.people = people;
-                });
+            // Just filter through our originalPeople array to find people who match the users search
+            this.people = this.originalPeople.filter(
```
OzzieOrca: Could we filter out people who are already in `$ctrl.assigned`? Although your current implementation seems to do it somehow... Does the ui-select filter out chosen ones somehow?

TheNoodleMoose: I think that is the right direction, the objects in assigned and the objects in people are different however, maybe we could filter by ID?

TheNoodleMoose:
I re-added the query and did some filtering to check if they are in this.assigned, which has seemed to fix this error. I do not know if its cleanest way to filter so if you have a better version please let me know.

app/assets/javascripts/angular/components/assignedPeopleSelect/assignedPeopleSelect.component.js
```diff
@@ -19,33 +19,39 @@ function assignedPeopleSelectController(
     RequestDeduper,
 ) {
     this.people = [];
-    this.originalPeople = [];
     this.isMe = assignedPeopleSelectService.isMe;
     this.$onInit = () => {
         const requestDeduper = new RequestDeduper();
         // Keep an original copy of all the people so we do not have to continually query
```
OzzieOrca: You can remove this comment.

app/assets/javascripts/angular/components/assignedPeopleSelect/assignedPeopleSelect.component.js
```diff
@@ -19,33 +19,39 @@ function assignedPeopleSelectController(
...
-                        .toLowerCase()
-                        .includes(search.toLowerCase()) ||
-                    person.last_name
-                        .toLowerCase()
-                        .includes(search.toLowerCase()),
-            );
+
+            assignedPeopleSelectService
+                .searchPeople(search, this.organizationId, requestDeduper)
+                .then(people => {
```
OzzieOrca: I think you can just do
```js
this.people = people.filter(person =>
  this.assigned.map(({ id }) => id).includes(person.id)
);
```
It's probably more efficient to precompute the array of assigned person ids instead of doing it in every loop.
```js
const assignedPeopleIds = this.assigned.map(({ id }) => id);
this.people = people.filter(person => assignedPeopleIds.includes(person.id));
```

It would be nice if this worked but I don't think the references are the same between the API data and the selected people... The references would break if the data is ever cloned so that might not be great.
```js
this.people = people.filter(person =>
  this.assigned.includes(person)
);
```
I think that's sorta what you were trying to fix by finding the people that match the assigned people by id and then filtering them by reference.

app/assets/javascripts/angular/components/assignedPeopleSelect/assignedPeopleSelect.component.js
```diff
@@ -19,33 +19,39 @@ function assignedPeopleSelectController(
...
+                    // Find all the current people in this.assigned
+                    const currentPeople = people.filter(person => {
+                        return this.assigned.some(assignedPerson => {
+                            return assignedPerson.id === person.id;
+                        });
+                    });
+                    // Map through all the people and if the currentPeople array does not include the person return that person
+                    // However it will produce undefined for a person if they are in current people so we then filter out all values that are undefined
+                    this.people = people
+                        .map(person => {
```
OzzieOrca: Just for future reference, I don't think you need the `map`.

```js
this.people = people
  .filter(person => currentPeople.includes(person))
```

TheNoodleMoose: Oh yes this makes sense, thanks!

app/assets/javascripts/angular/components/assignedPeopleSelect/assignedPeopleSelect.component.js
TheNoodleMoose: So precompute the assignedPeopleIds array on init instead of doing every loop makes sense. And then are you suggesting that assignedPeopleIds is bad idea because the reference would break if the data is cloned?  So should I keep the filtering how I have it currently? I've tried to use all these suggestions you made and it is currently working.  If it is a bad idea, I think keeping what I have and doing the first suggestion you made would be the best one, thanks for the suggestions :slightly_smiling_face:

OzzieOrca: `assignedPeopleIds` should work. You don't have to worry about references for simple data types like strings and numbers. Just objects and arrays.

I was trying to say if you filtered by doing `person1 === person2` I think it does that comparison by reference and if they aren't exactly the same object (even if all the data is the same) that might return false.

`person1.id === person2.id` will compare by value.

app/assets/javascripts/angular/components/assignedPeopleSelect/assignedPeopleSelect.component.js
```diff
@@ -21,6 +21,7 @@ function assignedPeopleSelectController(
     this.people = [];
     this.isMe = assignedPeopleSelectService.isMe;
     this.$onInit = () => {
+        const assignedPeopleIds = this.assigned.map(({ id }) => id);
```
OzzieOrca: I think this needs to be done in the `$watch` or the `.then` cuz `this.assigned` will change over time.

TheNoodleMoose: Yeah this makes sense, will move.

TheNoodleMoose:
I just realized that what I had was making this.people only be an array of people from this.assigned. And once no one was assigned the list would be completely empty and no search results would come up. So I changed it to now filter out everyone who is not assigned.

TheNoodleMoose:
So Alex was wondering if I could fix the alignment issue that occurs when you have multiple questions and you start adding selected items (images to see this here: https://jira.cru.org/browse/MHP-2764). How would I even go about solving this? Especially since you don't know how many items the user will have and the lengths of each item they select. If you think it would be better as its own ticket as well so we can get the auto labeling merged in let me know.

OzzieOrca:
Maybe a new ticket or at least PR would be good.

Screen Shot 2019-07-29 at 6 30 49 PM
Idk how it's currently laid out but you might be able to do something with flex and so the whole row expands when the input expands. For the example in the screenshot I guess you would need to start adding whitespace between the chocolate, sugar, and oatmeal inputs. Although the box around the right  side might make that harder...

TheNoodleMoose:
So with the new filtering I created, I guess it did not account for if this.assigned was null. This was then not allowing people to populate the selection field.

app/assets/javascripts/angular/components/assignedPeopleSelect/assignedPeopleSelect.component.js
```diff
@@ -33,11 +33,17 @@ function assignedPeopleSelectController(
...
                 .then(people => {
-                    const assignedPeopleIds = this.assigned.map(({ id }) => id);
-                    // Find all the current people in this.assigned
-                    this.people = people.filter(
-                        person => !assignedPeopleIds.includes(person.id),
-                    );
+                    if (this.assigned === null) {
+                        this.people = people;
+                    } else {
+                        const assignedPeopleIds = this.assigned.map(
```
OzzieOrca: I think you could just do this:
```suggestion
                        const assignedPeopleIds = (this.assigned || []).map(
```

Your solution runs less code if it's null which is good. Writing less code can make things easier to maintain. I'm fine with either and I'm approving what you've got.

PR #258, opened 2019-07-26
MHP-2742-Fix-Permission-Requires-Remove-JQuery-References

TheNoodleMoose:


app/assets/javascripts/angular/components/personProfile/personProfile.component.js
```diff
@@ -321,14 +320,15 @@ function personProfileController(
...
         if (hasEmailAddress) {
             vm.saveAttribute(vm.personTab.orgPermission, 'permission_id');
         } else {
             vm.personTab.orgPermission.permission_id = oldValue;
-            jQuery.a(
-                $filter('t')(
-                    'contacts.index.for_this_permission_email_is_required_no_name',
-                ),
+            // Use error service to displayError message
+            vm.personTab.orgPermission.$error = $filter('t')(
```
OzzieOrca: This isn't very dynamic but neither was the old version. It'll work but I don't think it'll take too much work to add a custom validator that will update whenever the email input changes. I was playing around with it and the error never goes away until you change the permission and you can delete the email but still have the permission and the frontend seems happy but the API throws an error.

You can
1. Make sure the form is initialized https://github.com/CruGlobal/give-web/blob/master/src/common/components/paymentMethods/creditCardForm/creditCardForm.component.js#L69-L74 https://github.com/CruGlobal/give-web/blob/master/src/common/components/paymentMethods/creditCardForm/creditCardForm.tpl.html#L1
2. Add a custom validator to the input (I think they are referenced by the name attribute). It's just a function that could go look up if the email field is filled/is valid. https://github.com/CruGlobal/give-web/blob/master/src/common/components/paymentMethods/creditCardForm/creditCardForm.component.js#L102
3. `<ng-messages>` might be overkill for displaying a single possibility but it's a nice way to render errors when `$error` contains a bunch of keys. https://github.com/CruGlobal/give-web/blob/master/src/common/components/paymentMethods/creditCardForm/creditCardForm.tpl.html#L67-L70
4. You might have to tell the permissions input to revalidate when the email field changes. https://github.com/CruGlobal/give-web/blob/master/src/common/components/paymentMethods/creditCardForm/creditCardForm.component.js#L108-L110

Let me know if this seems like it'll take a bit of time. We could just live with what we have.

TheNoodleMoose: I'll try to make it better by trying to recreate the solution you have suggested here. And when you delete the email, does that users org permissions need to be wiped as well?

OzzieOrca: Nah they should have to hit the x button. I guess it will just be invalid if it's empty but not deleted. Idk if that's straight forward or not...

I don't think this ticket is super important so if that feels like it will take more than an hour or 2, maybe just go with something simpler like what you have...

TheNoodleMoose: Yeah I agree, I think this will take me a while to get setup. I'm going to push up a change that sets the error to null once we run the function that adds their email so the error goes away once they add the email.

OzzieOrca:
Can you remove these too?

The API service should show a toast here on error so idk why there was separate jQuery handling...
https://github.com/CruGlobal/missionhub-web/blob/e5eb0a8895c05a1aa682df4cb68df11eb787d191/app/assets/javascripts/angular/components/peopleViewPerson/peopleViewPerson.component.js#L114
This will become unused after you've deleted the usages. https://github.com/CruGlobal/missionhub-web/blob/e5eb0a8895c05a1aa682df4cb68df11eb787d191/app/assets/javascripts/angular/missionhubApp.constants.js#L11

TheNoodleMoose:
The first one I removed already and the second one I will remove :+1:

app/assets/javascripts/angular/components/personProfile/personProfile.html
```diff
@@ -125,7 +125,7 @@
                 ng-model="$ctrl.personTab.orgPermission.permission_id"
                 ng-change="$ctrl.permissionChange({{
                     $ctrl.personTab.orgPermission.permission_id
-                }})"
+                }}, $ctrl.personTab.orgPermission.permission_id)"
```
OzzieOrca: This is really weird but I guess it works. Does it need single quotes around the `{{expression}}`? https://stackoverflow.com/questions/26621104/ng-change-get-new-value-and-original-value

TheNoodleMoose: Yeah it is a little weird I thought as well. Based off QA you should be able to change the permission type to No permissions if they don't have an email associated to the person. And the single quotes seems to break certain aspects of the code so I think it's fine without it.

app/assets/javascripts/angular/components/personProfile/personProfile.scss
```diff
@@ -7,6 +7,11 @@ person-profile {
 
     display: block;
 
+    .unvalidated {
```
OzzieOrca:
```suggestion
    .invalid {
```
? Or `.error`. Idk this is a trivial comment. `unvalidated` implies you never checked it but then how did you know to make it red?

TheNoodleMoose: No you're right this makes sense, will change :+1:

PR #264, opened 2019-07-29
MHP-2797-Better-Messaging-For-Web-Users-With-No-Orgs

TheNoodleMoose:
Add check to initial peopleDashboard load to determine whether the user has organizational permissions or not. When the user takes the route has described(https://jira.cru.org/browse/MHP-2797), we check if the user has organizational permissions. If a user does not, that means they are a mobile user and we should redirect them to a page describing how to download the app.


app/assets/javascripts/angular/components/myPeopleDashboard/myPeopleDashboard.component.js
```diff
@@ -41,7 +42,16 @@ function myPeopleDashboardController(
...
-        loadAndSyncData();
+        await loadAndSyncData();
+
+        if (
+            // If the logged in person has no organization permissions, redirect the user to download mobile app
+            loggedInPerson.person.organizational_permissions.length === 0
+        ) {
+            // This page we redirect to will change
+            $window.location.href =
+                'http://get.missionhub.com/account-verification/';
```
OzzieOrca:
```suggestion
                'https://get.missionhub.com/account-verification/';
```

app/assets/javascripts/angular/components/myPeopleDashboard/myPeopleDashboard.component.js
```diff
@@ -41,7 +42,15 @@ function myPeopleDashboardController(
...
     vm.$onInit = async () => {
         await loggedInPerson.loadOnce();
-        loadAndSyncData();
+        await loadAndSyncData();
+
+        if (
+            // If the logged in person has no organization permissions, redirect the user to download mobile app
+            loggedInPerson.person.organizational_permissions.length === 0
+        ) {
+            // This page we redirect to will change
```
OzzieOrca: @TheNoodleMoose Can you remove this comment? This has been updated.

PR #272, opened 2019-07-30
MHP-2381-Remove-Bind-Dynamic-Directive

TheNoodleMoose:


app/assets/javascripts/angular/components/myPeopleDashboard/myPeopleDashboard.html
```diff
@@ -1,10 +1,10 @@
 <div class="my-people-dashboard">
     <h1 ng-show="$ctrl.loading" ng-click="$ctrl.retry">
-        {{'dashboard.loading_contacts' | t}}
+        {{ 'dashboard.loading_contacts' | t }}
     </h1>
     <div ng-if="$ctrl.noPeople" class="no-people-message ng-cloak">
         <h1>{{ $ctrl.noPeopleWelcome }}</h1>
-        <p bind-dynamic="'dashboard.no_contacts_help' | t"></p>
+        <p bind-dynamic="$ctrl.noContactsHelp"></p>
```
OzzieOrca: All these should use `ng-bind-html` and you should  delete the `bind-dynamic` directive from the codebase.

app/assets/javascripts/angular/components/myPeopleDashboard/myPeopleDashboard.component.js
```diff
@@ -39,6 +41,14 @@ function myPeopleDashboardController(
             userPreferencesService.organizationOrderChange(vm.organizations),
     };
 
+    this.noContactsHelp = $sce.trustAsHtml(
+        $filter('t')('dashboard.no_contacts_help'),
```
OzzieOrca: Sorry if I communicated the opposite with my code example. I was just looking for existing sce code. I think I'd prefer to import `i18next.t` directly. Right now I guess  it doesn't matter what you use now as the filter just calls `i18next.t`. Thinking about migrating to React in the future would use it directly but the components would need a huge rewrite anyways.

TheNoodleMoose: Good point will change to this.

app/assets/javascripts/angular/components/myPeopleDashboard/myPeopleDashboard.html
TheNoodleMoose: The only issue is the text shows up but the icons no longer show up that are apart of the translation. The bind dynamic is allowing the icons to populate the html. Changing to ng-bind-html seems to not allow the icons. I may just go through the process of adding everything directly into the component. Does the text of these still need to be translated?

OzzieOrca: Oh, the icons were probably the reason this was run through `$compile` in the first place. Ya if it seems simple enough, it might be better to just get them out of the translations so we don't have to worry about if they are escaped or compiled. Ya the text should still be translated.

TheNoodleMoose:
I moved the icons out and deleted the the bind directive. I didn't know if there is a better way to do the translation, since the icon needs to be in between the text so I had to split it into two parts. If you see a better option let me know.

app/assets/javascripts/angular/components/myPeopleDashboard/myPeopleDashboard.html
```diff
@@ -1,10 +1,14 @@
 <div class="my-people-dashboard">
     <h1 ng-show="$ctrl.loading" ng-click="$ctrl.retry">
-        {{'dashboard.loading_contacts' | t}}
+        {{ 'dashboard.loading_contacts' | t }}
     </h1>
     <div ng-if="$ctrl.noPeople" class="no-people-message ng-cloak">
         <h1>{{ $ctrl.noPeopleWelcome }}</h1>
-        <p bind-dynamic="'dashboard.no_contacts_help' | t"></p>
+        <span>
+            <span ng-bind-html="$ctrl.noContactsHelpPart1"></span>
```
OzzieOrca: If there's no HTML, you don't need the `ng-bind-html` and can just translate normally.

app/assets/javascripts/angular/components/myPeopleDashboard/myPeopleDashboard.html
```diff
@@ -1,10 +1,14 @@
...
+            <ng-md-icon icon="addContact"></ng-md-icon>
+            <span ng-bind-html="$ctrl.noContactsHelpPart2"></span>
```
OzzieOrca: I think you could get rid of the `<br>` here by using CSS.
Screen Shot 2019-07-30 at 2 28 27 PM
I added the `flex flex-column items-center` and `measure-wide` classes from https://tachyons.io

TheNoodleMoose: Oh sweet, thanks for sharing this. I was wondering where all these classes were coming from haha.

app/assets/javascripts/angular/components/organizationCleanup/organizationCleanup.component.js
```diff
@@ -17,11 +19,19 @@ function organizationCleanupController(
...
+    $sce,
 ) {
     this.archiveBeforeDate = new Date();
     this.archiveInactivityDate = new Date();
     this.checkIcon = checkIcon;
 
+    this.ministriesCleanupDescriptionPart1 = $sce.trustAsHtml(
+        i18next.t('ministries.cleanup.description.part1'),
+    );
+    this.ministriesCleanupDescriptionPart2 = $sce.trustAsHtml(
```
OzzieOrca: If you're rendering this without `ng-bind-html`, you don't need `$sce.trustAsHtml` and can just do the translation in the template.

app/assets/javascripts/angular/components/organizationCleanup/organizationCleanup.component.js
```diff
@@ -17,11 +19,19 @@ function organizationCleanupController(
     $filter,
     $uibModal,
     $state,
+    $sce,
 ) {
     this.archiveBeforeDate = new Date();
     this.archiveInactivityDate = new Date();
     this.checkIcon = checkIcon;
 
+    this.ministriesCleanupDescriptionPart1 = $sce.trustAsHtml(
```
OzzieOrca: Same here, you don't need `$sce.trustAsHtml`.

app/assets/javascripts/angular/components/myPeopleDashboard/myPeopleDashboard.component.js
```diff
@@ -50,6 +52,19 @@ function myPeopleDashboardController(
...
+        this.noContactsHelpPart1 = $sce.trustAsHtml(
+            i18next.t('dashboard.no_contacts_help.part1'),
+        );
+        this.noContactsHelpPart2 = $sce.trustAsHtml(
+            i18next.t('dashboard.no_contacts_help.part2'),
+        );
+        this.editOrgOrderHelpPart1 = $sce.trustAsHtml(
+            i18next.t('dashboard.edit_org_order_help.part1'),
+        );
+        this.editOrgOrderHelpPart2 = $sce.trustAsHtml(
```
OzzieOrca: Same here, you don't need `$sce.trustAsHtml`.

app/assets/javascripts/angular/components/myPeopleDashboard/myPeopleDashboard.component.js
```diff
@@ -50,6 +52,19 @@ function myPeopleDashboardController(
...
             name: loggedInPerson.person.first_name.toUpperCase(),
         });
+        // Had to Split into two parts due to icon needings to be in between
+        this.noContactsHelpPart1 = $sce.trustAsHtml(
+            i18next.t('dashboard.no_contacts_help.part1'),
+        );
+        this.noContactsHelpPart2 = $sce.trustAsHtml(
+            i18next.t('dashboard.no_contacts_help.part2'),
+        );
+        this.editOrgOrderHelpPart1 = $sce.trustAsHtml(
```
OzzieOrca: Same here, you don't need `$sce.trustAsHtml`.

app/assets/javascripts/angular/components/myPeopleDashboard/myPeopleDashboard.component.js
```diff
@@ -50,6 +52,19 @@ function myPeopleDashboardController(
         vm.noPeopleWelcome = i18next.t('dashboard.no_contacts.welcome', {
             name: loggedInPerson.person.first_name.toUpperCase(),
         });
+        // Had to Split into two parts due to icon needings to be in between
+        this.noContactsHelpPart1 = $sce.trustAsHtml(
+            i18next.t('dashboard.no_contacts_help.part1'),
+        );
+        this.noContactsHelpPart2 = $sce.trustAsHtml(
```
OzzieOrca: Same here, you don't need `$sce.trustAsHtml`.

app/assets/javascripts/angular/components/myPeopleDashboard/myPeopleDashboard.component.js
```diff
@@ -50,6 +52,19 @@ function myPeopleDashboardController(
         vm.noPeopleWelcome = i18next.t('dashboard.no_contacts.welcome', {
             name: loggedInPerson.person.first_name.toUpperCase(),
         });
+        // Had to Split into two parts due to icon needings to be in between
+        this.noContactsHelpPart1 = $sce.trustAsHtml(
```
OzzieOrca: Same here, you don't need `$sce.trustAsHtml`.

app/assets/javascripts/angular/components/organizationCleanup/organizationCleanup.component.js
```diff
@@ -1,3 +1,5 @@
+import i18next from 'i18next';
```
OzzieOrca: This import still used?

TheNoodleMoose: It currently is not, but there are some translations using $filter in this file. Should I switch this over to use i18next directly?

OzzieOrca: I'm fine with either switching them or just deleting the import if it's unused.

app/assets/javascripts/angular/components/myPeopleDashboard/myPeopleDashboard.html
```diff
@@ -1,10 +1,17 @@
...
-    <div ng-if="$ctrl.noPeople" class="no-people-message ng-cloak">
+    <div
+        ng-if="$ctrl.noPeople"
+        class="no-people-message ng-cloak flex flex-column items-center"
+    >
         <h1>{{ $ctrl.noPeopleWelcome }}</h1>
-        <p bind-dynamic="'dashboard.no_contacts_help' | t"></p>
+        <span class="measure-wide">
+            <span>{{ 'dashboard.no_contacts_help.part1' | t }}</span>
+            <ng-md-icon icon="addContact"></ng-md-icon>
```
OzzieOrca: I don't know that your changes did this, but the question icon is not the `addContact` icon. Just noticed this, guess I wasn't thinking too hard about it before. It might be showing the question mark cuz it can't find the requested icon...
Screen Shot 2019-08-02 at 11 43 57 AM
Is that icon missing for some reason?

TheNoodleMoose: Yeah I went to one of the other Netlify builds and they all have this missing symbol. I'll look into this.

OzzieOrca: Thanks :) You can make a ticket if it isn't a trivial solution.

TheNoodleMoose: Is there documentation for ng-md-icons? Is it similar to font-awesome icons where you can only use icons in their library?  I found this https://klarsys.github.io/angular-material-icons/, but can't find icons like editOrder, which is working. I changed it from addContact to person_add and got this.
Screen Shot 2019-08-02 at 3 20 55 PM

OzzieOrca: It's added as a custom icon. https://github.com/CruGlobal/missionhub-web/blob/12b1dee91cb11fead4c38c6366b3e9ed964ba665/app/assets/javascripts/angular/missionhubApp.config.js#L109 Seems like a really weird way of doing things... We'll do something different as we move away from Angular.

OzzieOrca: But that looks better than the question mark :)

Looks  like it may be missing `<svg>` off the front...

TheNoodleMoose: Interesting, that is a strange way of doing it. Think i figured out the issue. It was under addContact but I changed it to addPerson and got this: 
Screen Shot 2019-08-02 at 3 32 26 PM

Is this the correct icon?

OzzieOrca: Probably. You can ask Alex but I don't know of other versions...

TheNoodleMoose: Sweet will do, thanks Scotty :)

PR #280, opened 2019-08-01
MHP-2659: Dashboard Challenges

bwtoptal:


src/dashboard/containers/ChallengesPage/index.js
```diff
@@ -0,0 +1,105 @@
...
+
+        const range = (start, end) => {
+            const format = 'MM/DD/YYYY';
+            const startDate = moment(start).format(format);
+            const endDate = moment(end).format(format);
+            return `${startDate} - ${endDate}`;
+        };
+
+        const rows = [];
+        data.globalCommunityChallenges.nodes.forEach(
```
OzzieOrca: I think you could do this with a `.map`.

```js
return data.globalCommunityChallenges.nodes.map(
  ({ id, title, acceptedCount, completedCount, createdAt, endDate }) => [
    id,
    title,
    acceptedCount,
    percentage(completedCount, acceptedCount),
    duration(createdAt, endDate),
    range(createdAt, endDate)
  ]
);
```

src/dashboard/containers/ChallengesPage/queries.js
```diff
@@ -0,0 +1,59 @@
...
+            impactReport {
+                interactionsCount
+                interactionsReceiversCount
+                challengesCount
+            }
+        }
+    }
+`;
+
+const GET_GLOBAL_CHALLENGES = gql`
```
OzzieOrca: Where are global challenges shown in Figma? I feel like I'm missing something...

src/dashboard/containers/ChallengesPage/index.js
bwtoptal: Haha... Thanks for that! Much better this way.. 👍

src/dashboard/containers/ChallengesPage/queries.js
bwtoptal: No, you are absolutely right, there is no global challenges page. I just had to use this endpoint, since the `organization.communityChallenges` didn't work for me and I wanted to implement pagination etc, so I used a query for global challenges which has similar response format. Carlos is looking into it.

OzzieOrca:
Bartosz said this was no longer needed. Must have been implemented elsewhere.

PR #308, opened 2019-08-08
MHP-2831-Fix-Alignment-On-Surveys

TheNoodleMoose:


TheNoodleMoose:
@OzzieOrca Any reason why the yarn test:react:coverage is failing? I have the latest commits from master which is passing so I'm not to sure why it is failing.

PR #313, opened 2019-08-09
MHP-2807-Fix-The-Unassigned-Count

TheNoodleMoose:
If a person is added or deleted I just emit _updateFilterCount_ and catch that emit in the peopleFiltersPanel component. If there is a better way for catching if an update occurs let me know.


app/assets/javascripts/angular/components/peopleFiltersPanel/peopleFiltersPanel.component.js
```diff
@@ -32,8 +32,6 @@ function peopleFiltersPanelController(
     this.filters = {};
 
     this.$onInit = () => {
-        loadFilterStats();
```
OzzieOrca: Do we still want this to run initially? Or does something else handle it?

app/assets/javascripts/angular/components/peopleFiltersPanel/peopleFiltersPanel.component.js
```diff
@@ -106,6 +108,8 @@ function peopleFiltersPanelController(
             };
         }
 
+        loadFilterStats();
```
OzzieOrca: Why do we need to do this here?

app/assets/javascripts/angular/components/peopleFiltersPanel/peopleFiltersPanel.component.js
TheNoodleMoose: Yeah, I will move it back into the init. It was running due to the other thing you commented on. It makes more sense to run it in the init.

app/assets/javascripts/angular/components/peopleFiltersPanel/peopleFiltersPanel.component.js
TheNoodleMoose: Yeah somewhere along the way it got moved here before I added the $rootScope catch for testing. I will move it from here back into the init.

PR #318, opened 2019-08-14
Refactor of question assignment layout

TheNoodleMoose:
This refactor works great, but I'm running into an issue with the ng-sortable not working properly. It seems that the answers are not actually changing position, so the when the onEnd function fires it resets the answers back to their original positions.


app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.html
```diff
@@ -230,88 +180,225 @@
...
-                                    disabled="rule.trigger_keywords === '' || !$ctrl.directAdminPrivileges"
-                                ></assigned-people-select>
-                                <a
-                                    href
-                                    class="ml3 delete"
-                                    ng-click="$ctrl.deleteQuestionContent(question, question.question_answers, rule, $index);"
-                                    ng-if="$ctrl.directAdminPrivileges"
+                                <div
+                                    class="relative flex question-row assignments pb2"
+                                    ng-repeat="rule in question.question_rules | filter: { rule_code: 'AUTOASSIGN' } track by $index"
```
OzzieOrca: I think I had issues with an ng-repeat before that had a filter and/or track by. I could try and look that commit up. I've got a meeting but I'll come back to this.

TheNoodleMoose: Sweet, yeah I'd appreciate it. Thanks Scotty :slightly_smiling_face:

OzzieOrca: Looks like track by is fine. I removed a limitTo filter in https://github.com/CruGlobal/missionhub-web/pull/151/files#diff-dcca7b7aa2235de0cda9286b6416d83cL8 to get it to work since it wouldn't work with it. Maybe the `filter` here is the same issue. We don't appear to be using it on other ng-sortables. You could test without in and if it works I guess we'll have to do that filtering in the controller or something...

Sorta makes sense. How is drag and drop going to reorder a partial list. Should the moved item go before or after the hidden items. Maybe there are sortable libraries that handle it better, idk.

TheNoodleMoose:
I switched it to repeat through the answers like it was originally, so that fixed it :)

TheNoodleMoose:
@OzzieOrca So I noticed that since displaying the assign-people/label-select is tied to whether rule.trigger_words === answer, if we add two questions in row without filing in the answers trigger word this happens: 
Screen Shot 2019-08-16 at 2 37 39 PM

Do you have any suggestions? This also happens if two questions are the same which could be an issue. I'm going to continue and see if there is a better option.

TheNoodleMoose:
So I moved from basing the ng-if off of the answer to the index of that answer that way even if there are two of the same answers, the indexes will never be the same. If there is a cleaner way of doing what I did here let me know!

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.html
```diff
@@ -230,90 +180,245 @@
...
+                                                />
+                                            </div>
+                                            <div>
+                                                ----
+                                            </div>
+                                        </div>
+                                        <div
+                                            class="relative flex pb2 w-48"
+                                            ng-if="(question.question_rules | filter: { rule_code: 'AUTOASSIGN' }).indexOf(rule) === question.question_answers.indexOf(answer)"
+                                            ng-repeat="rule in question.question_rules | filter: { rule_code: 'AUTOASSIGN' } track by $index"
```
OzzieOrca: If you only ever want to show one rule (which seems like what you're doing with the `ng-if`), do you need the `ng-repeat`? Can you just reference the correct rule when you need it?

TheNoodleMoose: Yeah this makes sense, so should I make some function in the controller that filters out all the rules based on the rule_type(either auto-assign or auto-notify) and then filter out all the other ones that don't match the index? Seems better to do it outside the template haha.

TheNoodleMoose: To answer my self, Scotty, you were correct. I can just reference the correct rule, no function needed. All I had to do was filter by AUTOASSIGN or AUTONOTIFY and then reference by the $index. Thanks Scotty :)

OzzieOrca:
Looks better. That's still a lot of array filtering. I assume AngularJS has to run each expression individually. You could do it once with https://docs.angularjs.org/api/ng/directive/ngInit (although using it is discouraged). Maybe you could add a property on the controller but I guess that could get out of date when answers are added/deleted.

TheNoodleMoose:
Yeah I agree, it is a ton of filtering but to get the correct index we have to filter some. I could try the ng-init if we think it'll be more optimal than all this filtering. Yeah the controller option seems unnecessary, I tried it but it seemed a little redundant and similar to what we are doing here.

TheNoodleMoose:
@OzzieOrca Do you think this is good to move into QA?

PR #321, opened 2019-08-14
MHP-2865-Fields-Boxes-Not-Showing

TheNoodleMoose:
Forms.scss was overriding all inputs with type of text to remove border. Just created a class to override that removal.


app/assets/javascripts/angular/components/orgManagement/orgManagementEdit.html
```diff
@@ -1,7 +1,7 @@
 <div class="accordion-body pb3 pt3">
     <h3 class="f3 fw6 pivot-dark-blue">
-        {{$ctrl.new ? 'common:organizations.new' : 'common:organizations.edit' |
-        t}}
+        {{ $ctrl.new ? 'common:organizations.new' : ('common:organizations.edit'
```
OzzieOrca: Is this prettier doing this? Did these files somehow not get run through it?

TheNoodleMoose: Yeah prettier seems to do this randomly on what seems to be already formatted code. I think it maybe my settings, I'm not too sure.

TheNoodleMoose:
@OzzieOrca So it seems that forms.scss is supposed to be a general file for forms defaults as most inputs are using it in some capacity. Did the colors/defaults change recently across the project and the defaults just have not been updated? If so we could make a ticket and I could go through and properly update the defaults which would hopefully solve the random css issues.

OzzieOrca:
Some of the blues might have changed... Might be good to try and update it.

PR #335, opened 2019-08-20
MHP-2882-Ng-Duplicates-Rollbar-Error

TheNoodleMoose:
I added the track by $index to these two repeats to remove a ton of the rollbar errors we were receiving. Upon originally adding this however, I noticed once the email is added is produced to inputs with the same email or phone number in them. 
Screen Shot 2019-08-20 at 10 20 09 AM

I did some deeper diving and it seems the api response is producing two identical arrays with the same email address, id, etc. I'm not sure if this a deeper issue, and if its the api or client. So what I did was just create a filter that filters out the extra copy that is created.   

Here is the response from the api, as you can see the array is empty and then produces two copies. Everything seems to only run once, so I'm not too sure why this is occurring.

Screen Shot 2019-08-20 at 10 21 01 AM


TheNoodleMoose:
Yeah this really felt hacky and like I was circumventing the actual issue at hand. I'll see if I can mess with this a little more to resolve the actual issue.

TheNoodleMoose:
@OzzieOrca So upon further investigation I'm still somewhat confused on what's happening. 
Console logging vm.personTab.person before (I assume this is what is actually saving to the api):
https://github.com/CruGlobal/missionhub-web/blob/463dc74bad02669d5384b5fba65425f4640a4152/app/assets/javascripts/angular/components/personProfile/personProfile.component.js#L185-L198 

shows that there is a duplicate of the email inside of vm.personTab.person.email_addresses. 
Screen Shot 2019-08-21 at 5 34 03 PM
But when I log the JsonApiDataStore there is only one of that email so I'm not to sure where the second one is coming from as it is being added before we even save to the api.
I guess I need to figure out exactly where we are updating vm.personTab.person.email_addresses.

OzzieOrca:
Idk I could look at it tomorrow maybe... It's really hard to follow things that go through the JsonApiDataStore... I think when you add a new address, one gets added to the store without an id before the API request is sent to create that address and then the new one comes back and is also added. But I haven't debugged it to see if that's true. Idk why track by is complaining either as it should be tracking by id and the clientside one wouldn't have an id assigned by the server yet.

app/assets/javascripts/angular/components/personProfile/personProfile.service.js
```diff
@@ -123,7 +123,14 @@ function personProfileService(
                 {
                     included: httpProxy.includedFromModels(relationships),
                 },
-            );
+            ).then(res => {
```
TheNoodleMoose: After saving to the api, the response includes two of the added email/number here. So I just moved the filtering right here. I still can't figure out why the response it returning the duplicate. 

We are sending just one copy to the api as we should be:
Screen Shot 2019-08-26 at 8 56 45 AM

And in the response, the included shows the one added email/number: 
Screen Shot 2019-08-26 at 8 57 05 AM

But I noticed in the relationships there are two copies of the id there: 
Screen Shot 2019-08-26 at 8 57 27 AM

I'm not sure if this is of any significance or if it is contributing to the issue at all.

TheNoodleMoose:
@OzzieOrca I think we should add these additional filters until the api issue is resolved so we aren't wasting our rollbar usage. Once this gets fixed I can remove them. What do you think?

app/assets/javascripts/angular/components/personHistory/personHistory.component.js
```diff
@@ -104,7 +104,7 @@ function personHistoryController(
 
     // Scroll to the bottom of the history list
     function scrollToBottom() {
-        const scrollContainer = $document.getElementsByClassName(
+        const scrollContainer = $window.document.getElementsByClassName(
```
TheNoodleMoose: I know this is really random to be in here but noticed it coincided with errors like this: https://rollbar.com/Cru/missionhub-web/items/2782/. This will fix this rollbar issue: https://rollbar.com/Cru/missionhub-web/items/1783/

app/assets/javascripts/angular/components/organizationOverviewSurveys/organizationOverviewSurveys.component.js
```diff
@@ -42,8 +42,6 @@ function organizationOverviewSurveysController(
                 size: 'sm',
             })
             .result.then(newSurvey => {
-                this.organizationOverview.surveys.push(newSurvey);
```
OzzieOrca: Why isn't this needed? Does something else reload it?

app/assets/javascripts/angular/components/personProfile/personProfile.component.js
```diff
@@ -190,7 +191,7 @@ function personProfileController(
 
         $scope.$emit('personModified');
 
-        var savePromise = personProfileService.saveRelationship(
+        let savePromise = personProfileService.saveRelationship(
```
OzzieOrca:
```suggestion
        const savePromise = personProfileService.saveRelationship(
```

app/assets/javascripts/angular/components/personProfile/personProfile.component.js
```diff
@@ -222,8 +223,17 @@ function personProfileController(
         );
     }
 
+    function addressWithPending() {
+        let addresses = vm.personTab.person.addresses;
```
OzzieOrca:
```suggestion
        const addresses = vm.personTab.person.addresses;
```

app/assets/javascripts/angular/services/interactions.service.js
```diff
@@ -128,6 +128,17 @@ function interactionsService(
                 )
                 .then(httpProxy.extractModel)
                 .then(function(interaction) {
+                    // Filter out extra initiator
+                    const filteredInteraction = interaction.initiators.filter(
+                        (initiator, index) => {
+                            return (
```
OzzieOrca: You don't need the `return` or `{}`. The arrow function can just return the value.

app/assets/javascripts/angular/components/personProfile/personProfile.service.js
```diff
@@ -123,7 +123,14 @@ function personProfileService(
...
+                // Filter out extra response
+                return (person[relationshipName] = res.data[
+                    relationshipName
+                ].filter((data, index) => {
+                    return res.data[relationshipName].indexOf(data) >= index;
```
OzzieOrca: Don't need `return`.

app/assets/javascripts/angular/components/personProfile/personProfile.service.js
```diff
@@ -123,7 +123,14 @@ function personProfileService(
                 {
                     included: httpProxy.includedFromModels(relationships),
                 },
-            );
+            ).then(res => {
+                // Filter out extra response
+                return (person[relationshipName] = res.data[
```
OzzieOrca: Don't need `return`.

app/assets/javascripts/angular/components/personProfile/personProfile.component.js
```diff
@@ -222,8 +223,17 @@ function personProfileController(
...
+        const filteredAddresses = addresses.filter((address, index) => {
+            return addresses.indexOf(address) >= index;
```
OzzieOrca: Don't need `return`;

app/assets/javascripts/angular/components/organizationOverviewSurveys/organizationOverviewSurveys.component.js
TheNoodleMoose: When a survey is created it redirects to that new page to update the new survey. When we go back to the page of surveys, we do a new GET request so it is automatically updated and we don't need to push this extra copy. Removing this seems not to change anything, except removing this error.

TheNoodleMoose:
@OzzieOrca I removed the filtering for phone number, email, and addresses as https://jira.cru.org/browse/MHP-2907 fixed it. However, Spencer said that the duplicate interactions was not fixed so I left it in.

PR #336, opened 2019-08-20
MHP-2879-Remove-Help-Beacon-From-Surveys

TheNoodleMoose:


index.ejs
```diff
@@ -1,37 +1,96 @@
...
-            topArticles: true
-        });
-    </script>
-
-</body>
+        <script>
+            // Disable the help scout beacon if the user is taking a survey
+            if (
+                window.location.origin !== 'https://stage.mhub.cc' ||
+                'https://mhub.cc'
```
OzzieOrca:
```suggestion
                window.location.origin !== 'https://mhub.cc'
```

index.ejs
```diff
@@ -1,37 +1,96 @@
...
+                        c = e.createElement('script');
+                    (c.type = 'text/javascript'),
+                        (c.async = !0),
+                        (c.src = 'https://djtflbt20bdde.cloudfront.net/'),
+                        r.parentNode.insertBefore(c, r);
+                })(document, window.HSCW || {}, window.HS || {});
+            }
+        </script>
+        <script>
+            if (
```
OzzieOrca: Maybe you could combine these script tags and only have one `if` statement.

index.ejs
TheNoodleMoose: :+1:

index.ejs
TheNoodleMoose: Will do!

index.ejs
```diff
@@ -1,37 +1,90 @@
...
-    r.parentNode.insertBefore(c,r)}(document,window.HSCW||{},window.HS||{});</script>
-    <script>
-        HS.beacon.config({
-            topArticles: true
-        });
-    </script>
+        <script>
+            // Disable the help scout beacon if the user is taking a survey
+            if (
+                window.location.origin !== 'https://stage.mhub.cc' ||
```
OzzieOrca: Staring at this again... I think you want
```js
window.location.origin !== 'https://stage.mhub.cc' && window.location.origin !== 'https://mhub.cc'
```
or
```js
!(window.location.origin === 'https://stage.mhub.cc' || window.location.origin === 'https://mhub.cc')
```
which are the same thing because of De Morgan's laws.

What you have right now will always be true.

TheNoodleMoose: Oh yeah that does make sense looking at, nice catch!

TheNoodleMoose:
Would this be better to merge to staging first, just to make sure it is working as intended and then merge to master?

TheNoodleMoose:
So the Netlify link should reflect the changes made in this commit correct? Or will this need to be merged to staging first and then tested there? I couldn't test this locally because you cannot open the surveys on localhost. But testing through the netlify link, the HS Beacon is still there when filling out a survey so this may not be the solution.

OzzieOrca:
Surveys should work on Netlify/localhost https://deploy-preview-336--missionhub.netlify.com/s/19130 but ya your domain check won't do anything until it's actually hosted there.

Ya you could merge to staging for QA.

TheNoodleMoose:
Everything works on staging :+1:

PR #342, opened 2019-08-20
MHP-2886: Limit number of reports requested

OzzieOrca:


app/assets/javascripts/angular/components/myPeopleDashboard/myPeopleDashboard.component.js
```diff
@@ -85,11 +85,11 @@ function myPeopleDashboardController(
     }
 
     function loadReports() {
-        const people = JsonApiDataStore.store.findAll('person');
+        const people = vm.organizations.flatMap(({ people }) => people);
```
TheNoodleMoose: Looks good :) Do you think moving away from JsonApiDataStore will be better in the long run for other parts of the application? Similar to the issues being experienced in https://github.com/CruGlobal/missionhub-web/pull/335. It seems like JsonApiDataStore can be somewhat of a catch all which may not be ideal..

OzzieOrca: I hate `JsonApiDataStore`. It works ok as a deserialization library but it should never be used as a caching library. It makes the data flow so hard to follow and finding all of a type regardless of how it was queried isn't useful and just causes unexpected behavior.

app/assets/javascripts/angular/services/person.service.js
```diff
@@ -290,15 +293,34 @@ function personService(
...
                             : organizationId,
-                        'page[limit]': 1000,
+                        'page[limit]': pageLimit,
                     },
                     {
                         errorMessage:
                             'error.messages.person.get_contact_assignments',
                     },
                 )
-                .then(httpProxy.extractModels)
```
TheNoodleMoose: What exactly does the httpProxy.extractModels do? I see it in some places and assume it's getting the response from the query in some capacity.

OzzieOrca: This https://github.com/CruGlobal/missionhub-web/blob/master/app/assets/javascripts/angular/services/httpProxy.service.js#L21. Same thing as https://github.com/CruGlobal/missionhub-web/pull/342/files/fa1a609237772c7e825923b313d65c37ae88f307..d0b5e9b3aeb76c09b171c9fa2d480b008dc82e9c#diff-2fa7c07e7185cab1a78433763082fee3R304 but I wanted to pull out the meta total and log it.

PR #356, opened 2019-09-05
Cleanup emotion theme typings

OzzieOrca:


OzzieOrca:
Yep. Feels a little hacky but I didn't find anything better. https://emotion.sh/docs/typescript#define-a-theme didn't work and requires you to import the custom module everywhere instead.

PR #357, opened 2019-09-05
Run lint on travis

OzzieOrca:


.travis.yml
```diff
@@ -10,6 +10,7 @@ script:
     - yarn gql:schema
     - yarn gql:codegen
     - yarn test:react:coverage
+    - yarn lint
```
TheNoodleMoose: So with this we are linting both js and ts, so do we need the `yarn lint:ts` anymore?

OzzieOrca: I wish they were combined more. `yarn lint` is running eslint. `yarn lint:ts` is running the typescript compiler (sorta as a linter). I think there was one plugin to integrate them but it wasn't working. Babel is still compiling all the code, it just throws away all the types and doesn't check them so that's why we're running the typescript compiler as a linter.

PR #369, opened 2019-09-06
MHP-2916-Update-Nav-For-Web

TheNoodleMoose:
first-child and nth-child were giving an error about server side rendering issues, which wasn't affecting us but changing to the suggested type works just fine and eliminated the error.

### DIFF
--- a/app/assets/javascripts/angular/components/organizationOverview/organizationOverview.html
+++ b/app/assets/javascripts/angular/components/organizationOverview/organizationOverview.html
@@ -39,9 +39,11 @@
         </div>
 
         <div class="tabs">
-            <div class="tab pivot-white">
+            <div
+                class="tab pivot-white"
+                ng-if="!$ctrl.envService.is('production')"
+            >
                 <a
-                    ng-if="!$ctrl.envService.is('production')"
                     ng-href="/ministries/{{ $ctrl.org.id }}/insights/personal"
                     ui-state-params="{ orgId: $ctrl.org.id }"
                     >{{ 'nav.insights' | t }}</a
--- a/app/assets/javascripts/angular/components/organizationOverview/organizationOverview.html
+++ b/app/assets/javascripts/angular/components/organizationOverview/organizationOverview.html
@@ -12,7 +12,7 @@
                 class="pivot-darker-blue f3 ph3 pt3 "
                 ui-sref="^.ministry.defaultTab({ orgId: $ctrl.org.id })"
             >
-                {{$ctrl.org.name}}
+                {{ $ctrl.org.name }}
             </div>
             <div ng-if="$ctrl.editMode">
                 <ng-md-icon
@@ -39,6 +39,13 @@
         </div>
 
         <div class="tabs">
+            <div class="tab pivot-white">
+                <a
+                    ng-href="/ministries/{{ $ctrl.org.id }}/insights/personal"
+                    ui-state-params="{ orgId: $ctrl.org.id }"
+                    >{{ 'nav.insights' | t }}</a
+                >
+            </div>
             <div
                 class="tab pivot-white"
                 ng-repeat="tabName in $ctrl.tabNames"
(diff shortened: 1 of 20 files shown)


app/assets/javascripts/angular/components/organizationOverview/organizationOverview.html
```diff
@@ -39,6 +39,13 @@
         </div>
 
         <div class="tabs">
+            <div class="tab pivot-white">
+                <a
+                    ng-href="/ministries/{{ $ctrl.org.id }}/insights/personal"
```
OzzieOrca: I don't think we're quite ready to have this on production yet (the GraphQL API isn't deployed there yet). Maybe you could add a `!envService.is('production')` check to hide this from prod for now.

app/assets/javascripts/angular/components/organizationOverview/organizationOverview.html
```diff
@@ -39,6 +39,14 @@
         </div>
 
         <div class="tabs">
+            <div class="tab pivot-white">
+                <a
+                    ng-if="!envService.is(production)"
```
OzzieOrca: Did you test this? You can force a specific environment https://yarnpkg.com/en/package/angular-environment#set-string-environment- or you could change the domains https://github.com/CruGlobal/missionhub-web/blob/76a596ca049c5c7b4895c4abba660060f4bdb691/app/assets/javascripts/angular/missionhubApp.config.js#L5-L19.

I'd be surprised if this worked as is. I think you have to inject `envService` into the controller and then somehow pass it or the return value of `is` to the template.

TheNoodleMoose: I was curious how it was working but when I got rid of the exclamation it disappeared. But injecting the envService makes more sense and guarantees it'll work so I'll do that.

app/assets/javascripts/angular/components/organizationOverview/organizationOverview.html
```diff
@@ -39,6 +39,14 @@
         </div>
 
         <div class="tabs">
+            <div class="tab pivot-white">
+                <a
+                    ng-if="!$ctrl.envService.is('production')"
```
OzzieOrca: This needs to be moved up to the div. The top is with the extra div deleted.
Screen Shot 2019-09-06 at 12 50 39 PM

TheNoodleMoose: Ahh, good point will change!

app/assets/javascripts/angular/services/interactions.service.js
```diff
@@ -23,7 +23,7 @@ function interactionsService(
             icon: 'note',
             title: 'application.interaction_types.comment',
             anonymous: false,
-            deprecated: false,
+            deprecated: true,
```
OzzieOrca: Can you give me some context on this change? I think comments still need to show up in the Angular UI. I haven't followed the code.

TheNoodleMoose: Alex wants to remove the icon from ministries and team page, but you are correct other components rely on this so I will add it back and find a different way of doing that.

TheNoodleMoose: Alex wants the note icon removed from the ministries and team page but you are correct, other components rely on it still. I will add it back and add a different method to remove it!

app/assets/javascripts/angular/components/organizationOverview/organizationOverview.html
```diff
@@ -1,15 +1,14 @@
 <async-content ready="$ctrl.ready">
     <div
-        class="br3 bg-pivot-darker-blue mb3"
+        class="br3 bg-pivot-dark-blue mb3"
         ng-class="{'hidden-org': $ctrl.org.visible === false, 'shadow-4': !$ctrl.isInsightsTab(), 'insights-tab': $ctrl.isInsightsTab()}"
         ng-if="$ctrl.showOrgNav()"
     >
         <div
-            class="flex items-center justify-between br3 br--top pb3"
-            ng-class="$ctrl.org.visible ? 'bg-pivot-white' : 'bg-pivot-gray-6'"
```
TheNoodleMoose: Also I'm really confused on what this does? If I click on an organization from the root menu, that organization opens up and the background is white. but if I refresh the page it is now grey. What exactly did this signal? I just removed it because on the design it is just supposed to be the blue at all times it seems.

OzzieOrca: I think at one point the colors on the main ministry page were different than once you navigated into one... Idk why... Deleting it sounds great.

PR #370, opened 2019-09-06
Add jest-emotion serializer

OzzieOrca:


OzzieOrca:
It kinda just looks like it counts top to bottom. So if you only made CSS changes, I assume it would just update rules for that element.

When you reorganize HTML, it might switch them on you. Not sure.

PR #372, opened 2019-09-06
MHP-2915-Updates-To-Dashboard

TheNoodleMoose:
I lowered the font-size by 25% and I'm going to try to begin moving some of these any types to being typed.


TheNoodleMoose:
The icon file names were on the wrong icons so I switched them around to be the file name that corresponds with it's correct icon.

src/communityInsights/components/StagesSummary/index.tsx
```diff
@@ -116,19 +116,52 @@ const Footer = styled.div`
     margin-top: 28px;
 `;
 
+// INTERFACES
+interface stagesReport {
```
OzzieOrca: You should be able to find this interface and a bunch of others in the `__generated__` files.

TheNoodleMoose: Certain folders did not produce a __generated__ folder. Like StagesSummary for example did not produce one. Should I be importing the interfaces created in the other folders to this one where they are needed?

OzzieOrca: They should be generated whenever the query is defined with a GraphQL tag.
```ts
gql```
  query queryName{...}```
```

Bartosz was putting some queries in common folders. I'm not sure what's best. If they are only used for a specific screen, I think it would be best to put them with that screen. Ya you can import them into wherever you need them.

TheNoodleMoose:
@OzzieOrca Also I noticed that on the challenges page, the back page button is not working. Looking into the queries response, it seems that the `hasPreviousPage` is always false even when it should be true. I don't know if that's something that still being worked on?

OzzieOrca:
Ya that's a bug... You want to make a new API ticket for it?

PR #373, opened 2019-09-11
MHP-2904-Web-org-team-tab-requests-all-previous-reports

TheNoodleMoose:


app/assets/javascripts/angular/components/organizationOverviewTeam/organizationOverviewTeam.component.js
```diff
@@ -42,11 +42,18 @@ function organizationOverviewTeamController(
         return organizationOverviewTeamService
             .loadOrgTeam(vm.organizationOverview.org, listLoader)
             .then(function(resp) {
-                vm.team = resp.list;
+                // Map through each person in the list and if they are not apart of vm.team already, push them to it
+                resp.list.map(person => {
```
OzzieOrca: Are you trying to deduplicate people here? Why? Could you just use `nextBatch` here too?
```ts
vm.team = [...vm.team, ...nextBatch];
```

Also if you want to deduplicate an array by reference you can do
```ts
const unique = [...new Set([1, 1, 2])];
```

TheNoodleMoose: I just pushed something that removed this mapping. For some reason I thought I needed to haha. I think doing `vm.team = resp.list;` will work just fine, the only thing that affects what is loaded is instead of using `resp.list`, use `resp.nextBatch` in the `loadTeamReports` function.

PR #376, opened 2019-09-17
Add GraphQL Playground as a lazy loaded dependency

OzzieOrca:


src/graphqlPlayground/GraphqlPlaygroundLazy.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { react2angular } from 'react2angular';
+
+import { GraphqlPlaygroundProps } from './GraphqlPlayground';
+
+const GraphqlPlayground = lazy(() =>
+    import(/* webpackChunkName: "graphql-playground" */ './GraphqlPlayground'),
+);
+
+const GraphqlPlaygroundLoader = (props: GraphqlPlaygroundProps) => (
+    <Suspense
```
TheNoodleMoose: So basically suspense allows this h1 to be rendered until the actual content we want is fully loaded? Super cool!

src/graphqlPlayground/GraphqlPlayground.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import { Provider } from 'react-redux';
```
TheNoodleMoose: Could we use useContext hooks here instead of redux? Or does the `graphql-playground-react` store only work with redux?

OzzieOrca: I think they used redux internally. I'm a little confused why the provider isn't also internal. Maybe they were having issues with multiple react-redux versions.

PR #377, opened 2019-09-19
Begin start of a NullState component, updated snapshots

TheNoodleMoose:
@OzzieOrca I began creating this component, tried to make it similar to how the react native application does it. Of course, this isn't finished as we have to wait for design. Just wanted to get your opinion on if this is a good way to handle this component and if there is any changes I should make.


src/communityInsights/components/NullState/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { withTheme } from 'emotion-theming';
+import { useTranslation } from 'react-i18next';
+
+interface Props {
+    impactInfo?: boolean;
+    width?: number;
+}
+
+const NullState = ({ impactInfo, width }: Props) => {
+    const LoadingContainer = styled.div`
```
OzzieOrca: This can live outside the functional component right? I think it would be executed less that way. You should be able to pass it props https://emotion.sh/docs/styled#changing-based-on-props.

src/communityInsights/components/NullState/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+            }
+            to {
+                background-color: ${({ theme }) => theme.colors.highlight};
+            }
+        }
+    `;
+    const { t } = useTranslation('insights');
+
+    return impactInfo ? (
+        <span>{t('gatheringStats')}</span>
```
OzzieOrca: Do you think it's useful to have a component to do this? Seems like it would be pretty easy to just change the text in the original component.

It's a little odd when components do 2 completely different things when you pass a different flag. If you think this would be good to have a shared component for, maybe just make a new component for it. Idk, what do you think?

TheNoodleMoose: *facepalms* yep, you are totally correct. Just another time where I was overthinking it lol. I'll change it back to have the text change inside the original component.

src/communityInsights/components/StagesSummary/index.tsx
```diff
@@ -156,17 +154,13 @@ const StagesSummary = ({ query, variables, mapData, longNames }: Props) => {
...
-                            <Stage key={entry.stage}>
-                                <Icon className={entry.icon} />
-                                <Title longNames={longNames}>
-                                    {entry.stage}
-                                </Title>
-                                <Value>{entry.count ? entry.count : '-'}</Value>
-                            </Stage>
-                        );
-                    })}
+                    {mapData(stepsOfFaithMockData).map((entry: any) => (
```
TheNoodleMoose: I made this use stepsOfFatihMockData instead of trying to use the NullState component.

src/communityInsights/containers/InteractionsPage/index.tsx
```diff
@@ -56,6 +56,7 @@ const InteractionsPage = () => {
                 subtitle={t('interactions.completedSubtitle')}
             >
                 <FiltersChart
+                    nullContent={'interactionsCompleted'}
```
TheNoodleMoose: Here is an example of passing the nullContent which is just the translation variable.

src/communityInsights/components/NullState/index.tsx
```diff
@@ -85,22 +92,53 @@ export const stepsOfFaithMockData = {
...
+            default:
+                return null;
+        }
+    };
+    return (
+        <LoadingContainer>
+            <img src={getImageSrc(content)}></img>
+            <LoadingText>{t(`nullState.${content}`)}</LoadingText>
+        </LoadingContainer>
+    );
```
TheNoodleMoose: @OzzieOrca Sorry for making this commit so large, with updating so many components and then updating snapshots everything kind of got out of hand lol. I'm wondering what you think of this? Basically I'm passing a nullContent variable which relates to that sections translation variable. I didn't know the best way to go about this over all. This seems to be a little much but to be able to dynamically get the needed image and translation this was what I came up with. Let me know if you have any suggestions 🙂

TheNoodleMoose:
@OzzieOrca Yeah i agree, I was only given the null states so i will talk to Alex to see if design has plans for a loading indicator.

src/communityInsights/components/BarChart/index.tsx
```diff
@@ -214,32 +209,33 @@ const BarChart = (props: Props) => {
...
-    //                         <LegendLabel>{legendLabel}</LegendLabel>
-    //                         <Line className={'dashed'} />
-    //                         <LegendLabel>{t('average')}</LegendLabel>
-    //                     </Legend>
-    //                 ) : null}
-    //             </Footer>
-    //         </div>
-    //     );
-    // }
+    if (nullCheck) {
```
TheNoodleMoose: I had to move the nullCheck here to allow the user to be able to change the date, as their month data may be null but year data could not be null.

src/communityInsights/components/FiltersChart/index.tsx
```diff
@@ -36,18 +36,22 @@ const FiltersChart = ({
...
-        return data.community.report.daysReport.filter((report: any) => {
-            return (
-                report.personalStepsAddedCount !== 0 &&
-                report.othersStepsAddedCount !== 0 &&
-                report.interactionsCount !== 0 &&
-                report.personalStepsCompletedCount !== 0 &&
-                report.othersStepsCompletedCount !== 0
-            );
-        });
+        const nullCheck = data.community.report.daysReport.filter(
```
TheNoodleMoose: Null check pretty much does the same thing but we just pass it to the barChart to determine what to render.

src/communityInsights/components/StepsChart/index.tsx
```diff
@@ -1,15 +1,18 @@
...
+import React from 'react';
 import styled from '@emotion/styled';
 
 import BarChart from '../BarChart';
 import NullState from '../NullState';
 
-const Wrapper = styled.div`
-    height: 400px;
+const Wrapper = styled.div<isNull>`
+    height: ${(props: isNull) => (props.isNull ? 'auto' : '400px')};
```
TheNoodleMoose: Had to do this weird isNull check here because without it the container for nullContent would be 400px. With just having the height set to auto it was slightly too large.

src/communityInsights/components/Table/__tests__/Table.tsx
```diff
@@ -26,35 +26,50 @@ describe('<Table />', () => {
         unmount();
     });
 
-    it('should render properly with data', async () => {
-        const mapRows = () => [['cell1', 'cell2'], ['cell3', 'cell4']];
-        const mapPage = () => ({});
+    // it('should render properly with data', async () => {
```
TheNoodleMoose: I had to disable this test, I'm not sure why is failing. The old test way using mock data for the old query so I updated it but its still not working correctly.

TheNoodleMoose:
@OzzieOrca So I simplified the logic so loading now just displays "loading...", and the null containers now only occurs for the StepsChart components and the challenges page. This way the user can still change the month or year on the charts.

app/assets/javascripts/angular/components/personProfile/personProfile.component.js
```diff
@@ -278,7 +278,6 @@ function personProfileController(
         );
         vm.pendingEmailAddress.setAttribute('email', '');
         vm.pendingEmailAddress.setAttribute('primary', false);
-        vm.pendingEmailAddress.setAttribute('$valid', true);
```
OzzieOrca: Looks like you've got some changes from that rollbar error PR. Were these intended?

package.json
```diff
@@ -107,10 +107,10 @@
         "@babel/plugin-transform-runtime": "^7.5.5",
         "@babel/preset-env": "^7.5.5",
         "@babel/runtime": "^7.4.3",
-        "@brainly/onesky-utils": "^1.3.2",
```
OzzieOrca: You must have resolved a merge wrong or something. You're downgrading versions here.

src/communityInsights/components/Table/__tests__/Table.tsx
OzzieOrca:
```suggestion
    xit('should render properly with data', async () => {
```
Does the same thing and it shows up in the Jest logs as having been skipped.

webpack.config.js
```diff
@@ -9,7 +9,7 @@ const BundleAnalyzerPlugin = require('webpack-bundle-analyzer')
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 const InlineManifestWebpackPlugin = require('inline-manifest-webpack-plugin');
 const FaviconsWebpackPlugin = require('favicons-webpack-plugin');
-const SriPlugin = require('webpack-subresource-integrity');
+// const SriPlugin = require('webpack-subresource-integrity');
```
OzzieOrca: This was reverted...

TheNoodleMoose:
@OzzieOrca I haven't merged from master in a while, do you think that should fix these issues?

OzzieOrca:
Idk, you could try. I'm confused.

TheNoodleMoose:
Honestly, me too. I'll see if I can figure it out.

src/communityInsights/components/BarChart/index.tsx
```diff
@@ -169,32 +175,65 @@ const BarChart = (props: Props) => {
         );
     };
 
+    const isNullCheck = (data: any) => {
```
TheNoodleMoose: So what I did here was move the nullCheck into the barChart so I could actually render a null Element into the svg of the barchart so we can still use the barChart if the content is null.

src/communityInsights/components/BarChart/index.tsx
```diff
@@ -169,32 +175,65 @@ const BarChart = (props: Props) => {
...
+            case 'interactionsCompleted':
+                return nullInteractions;
+            case 'challengesCompleted':
+                return nullChallenges;
+            default:
+                return null;
+        }
+    };
+
     const creatNull = () => {
```
TheNoodleMoose: Here is where I make the element that will be put into the barChart.

src/i18n/locales/en-US.ts
```diff
@@ -1171,24 +1171,46 @@ export default {
                     'community to open the MissionHub mobile app and complete a personal step of faith.',
             },
 
-            personalStepsAdded:
-                'This community does not have any members with personal steps of faith. Challenge your community to open the MissionHub mobile app and add a personal step of faith.',
-            communityMembersStages:
-                'This community does not have any members with stages. Challenge your current members to open the MissionHub mobile app and choose a stage.',
+            personalStepsAdded: {
```
TheNoodleMoose: I had to split the text up into two parts for the tspans.

src/communityInsights/containers/PersonalStepsPage/index.tsx
```diff
@@ -121,6 +122,8 @@ const PersonalStepsPage = () => {
                             (
                                 row: communityReportStagesPersonalStepsAddedRow,
                             ) => ({
+                                personalStepsOfFaith:
```
TheNoodleMoose: I had to add this in for the nullCheck, the one below it is translate as `Steps Of Faith` so I couldn't use it.

package.json
TheNoodleMoose: Yeah, just pulled from master and its updated now.

webpack.config.js
TheNoodleMoose: I reverted it back to the correct way. I am honestly not sure what happened and why it says there is not merge conflicts when there are.

TheNoodleMoose:
I'm just going to close this as it is too messy. A new one has been opened here: https://github.com/CruGlobal/missionhub-web/pull/416

PR #386, opened 2019-09-27
MHP-2818-Add-Test-To-Match-Datadog-Synthetic-Test

TheNoodleMoose:


cypress/integration/tests/missionHub-startup.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+        });
+    });
+});
+
+describe('People Dashboard', () => {
+    it('should load the people dashboard', () => {
+        cy.login();
+        cy.get('[href="/ministries/14665"]').should('be', 'Test Organization');
+    });
+});
```
mattdrees: Very cool!

Could I ask that you add a comment in this file somewhere noting that this mirrors a datadog synthetic test (this one: https://app.datadoghq.com/synthetics/details/u2s-kfb-d6j)? And maybe note that changes to this file probably need to be appropriately made to the synthetic test, as well.

TheNoodleMoose: Yes will do!

mattdrees: thanks!

.travis.yml
```diff
@@ -3,6 +3,9 @@ language: node_js
 node_js: '12'
 addons:
     chrome: stable
+    apt:
+        packages:
+            - libgconf-2-4
```
OzzieOrca: Can you put a comment about what is using this lib? I'm assuming it's Cypress but commenting about the dependency would help it not be accidentally be deleted or let someone know to delete it if Cypress is ever deleted.

cypress.json
```diff
@@ -0,0 +1,5 @@
+{
+    "baseUrl": "https://localhost:8080",
+    "chromeWebSecurity": false,
```
OzzieOrca: Is this still needed? Was this added when you were trying to navigate cross domain to The Key?

cypress/integration/tests/missionHub-startup.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+            .then(href => {
+                const keyUrl = href;
+                expect(keyUrl).to.contain(
+                    'https://thekey.me/cas/login?response_type=token&scope=fullticket&client_id=',
+                );
+            });
+    });
+});
+
+Cypress.Commands.add('login', () => {
```
OzzieOrca: Can you stick this in a new file if it's easy to do so? Or some way so that it could be shared with future tests?

Can you name this `signIn` to match the button name? I named it wrong initially... Trivial comment haha

cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
...
+// existing commands.
+//
+// For more comprehensive examples of custom
+// commands please read more here:
+// https://on.cypress.io/custom-commands
+// ***********************************************
+//
+//
+// -- This is a parent command --
+// Cypress.Commands.add("login", (email, password) => { ... })
```
OzzieOrca: Is this where we should put our `signIn` command?

cypress/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
```
OzzieOrca: Could this be shared with the rest of the app? Idk how Cypress works...

cypress.json
TheNoodleMoose: Yeah we added it then but for some reason it does not work without this set to false. Not too sure why.

cypress/tsconfig.json
TheNoodleMoose: I believe it could be but the official documentation recommends having its own tsconfig inside the cypress folder.

PR #387, opened 2019-10-03
Fix-Exception-Handler-Rollbar-Error

TheNoodleMoose:
Make sure that email address pending and phone number pending exist after deletion.


TheNoodleMoose:
https://github.com/CruGlobal/missionhub-web/blob/db5bf0910d1ee5eee12a444f60fecc6ee56b0533/app/assets/javascripts/angular/components/personProfile/personProfile.component.js#L254-L259

Right here is the culprit for the error. Basically after deletion, email_address_pending or phone_number_pending gets deleted and doesn't get created after. So when this check happens control doesn't exist.

app/assets/javascripts/angular/components/personProfile/personProfile.component.js
```diff
@@ -278,6 +278,7 @@ function personProfileController(
         );
         vm.pendingEmailAddress.setAttribute('email', '');
         vm.pendingEmailAddress.setAttribute('primary', false);
+        vm.pendingEmailAddress.setAttribute('$valid', true);
```
TheNoodleMoose: @OzzieOrca had to set this to true or after deletion it would always be invalid and would force the user to add either a phone number or email to remove the invalidation error.

PR #397, opened 2019-10-14
Responsive tabs

TheNoodleMoose:
WIP, things are broken still but getting prettier


app/assets/javascripts/angular/components/organizationOverview/organizationOverview.component.js
```diff
@@ -39,6 +37,12 @@ function organizationOverviewController(
         return $state.current.name === 'app.ministries.ministry.insights';
     };
 
+    this.navOpen = true;
```
OzzieOrca: This is named the opposite of what it represents. `orgNavCollapsed` maybe? Or go switch the logic everywhere.

PR #398, opened 2019-10-15
Make Org Nav Collapse On Click

TheNoodleMoose:
Also hid the icons when on the root menu


app/assets/javascripts/angular/components/organizationOverview/organizationOverview.html
```diff
@@ -96,43 +100,49 @@
                     <ul
                         class="dropdown-content br3 br--bottom shadow-5 overflow-hidden"
                     >
-                        <li>
+                        <li ng-click="$ctrl.orgNavOpen = false">
```
OzzieOrca: Should these be on the same tag as the ui-sref? Idk that it matters...

PR #409, opened 2019-10-17
Fix survey responsive icons

OzzieOrca:


OzzieOrca:
I made these styles more specific instead.

PR #416, opened 2019-10-23
MHP-2926-Create-Null-State-For-Dashboard

TheNoodleMoose:
I decided to make this new branch, because the other one was just a mess.


src/communityInsights/components/Table/index.tsx
```diff
@@ -146,6 +150,14 @@ const Table = ({ query, headers, mapRows, mapPage, variables }: Props) => {
         return <LoadingContainer>{t('loading')}</LoadingContainer>;
     }
 
+    const isNull = (data: any) => {
+        return data.community.communityChallenges.nodes.length <= 0;
+    };
+
+    if (isNull(data)) {
+        return <NullState content={nullContent} />;
+    }
```
TheNoodleMoose: The challenge page is the only one that does not use the bar chart so I had to use the NullState component here. It felt silly having a component(NullState) for only one component but I figured it may be used more in the future.

src/communityInsights/containers/PersonalStepsPage/index.tsx
```diff
@@ -114,12 +115,15 @@ const PersonalStepsPage = () => {
...
                 <StepsChart
+                    nullContent={'personalStepsAdded'}
                     query={GET_STAGES_REPORT_STEPS_ADDED}
                     mapData={(data: communityReportStagesPersonalStepsAdded) =>
                         data.community.report.stagesReport.map(
                             (
                                 row: communityReportStagesPersonalStepsAddedRow,
                             ) => ({
+                                personalStepsOfFaith:
+                                    row.personalStepsAddedCount,
```
TheNoodleMoose: I had to add this here for the null check in the barChart. The one below that is being used by the bar chart is being translated as `Personal Steps of Faith` so I couldn't use it to check if it was null.

src/communityInsights/containers/StepsOfFaithPage/index.tsx
```diff
@@ -112,10 +113,12 @@ const StepsOfFaithPage = () => {
                 subtitle={t('stepsOfFaith.addedSubtitle')}
             >
                 <StepsChart
+                    nullContent={'stepsOfFaithAdded'}
                     query={GET_STAGES_REPORT}
                     mapData={(data: communityReportStagesOthersStepsAdded) =>
                         data.community.report.stagesReport.map(
                             (row: communityStepsAddedStageReportRow) => ({
+                                stepsOfFaith: row.othersStepsAddedCount,
```
TheNoodleMoose: Same thing here, it was getting translated as well.

src/i18n/locales/en-US.ts
```diff
@@ -1161,6 +1161,56 @@ export default {
...
         monthLabel: '1M',
         yearLabel: '1Y',
+        gatheringStats: 'Gathering your stats...',
+        nullState: {
+            personalStepsCompleted: {
+                part1:
+                    'This community does not have any members who have completed a personal step of faith. Challenge your',
+                part2:
+                    'community to open the MissionHub mobile app and complete a personal step of faith.',
+            },
```
TheNoodleMoose: I had to break the sentences up in to two parts for the tspans in creatNull function.

src/communityInsights/components/BarChart/index.tsx
```diff
@@ -168,6 +171,74 @@ const BarChart = (props: Props) => {
...
+                        <tspan y="200" x="0" dy="1.6em" textAnchor="middle">
+                            {t(`nullState.${nullContent}.part2`)}
+                        </tspan>
+                    </text>
+                </g>
+            );
+        } else {
+            return null;
+        }
+    };
```
TheNoodleMoose: So after doing some thinking, I figured the best course of action in order to still be able to use the chart was to add a layer to the svg if the content is null. This allows the NullState to show up and the functionality of the chart to still be available, just like the design intended.

src/communityInsights/components/BarChart/index.tsx
```diff
@@ -168,6 +171,74 @@ const BarChart = (props: Props) => {
...
+            case 'interactionsCompleted':
+                return nullInteractions;
+            case 'challengesCompleted':
+                return nullChallenges;
+            default:
+                return null;
+        }
+    };
+    // Create a null element to be added as a layer in the barchart svg
+    const creatNull = () => {
```
OzzieOrca:
```suggestion
    const createNull = () => {
```

src/communityInsights/components/BarChart/index.tsx
OzzieOrca: Ok cool :) Thanks for figuring this out! 🎉

src/communityInsights/components/BarChart/index.tsx
OzzieOrca: No, you spelled `creat` wrong. 😆

OzzieOrca: You can keep the comment.

TheNoodleMoose: Oh my goodness how did I miss that! 🤣 🤦‍♂ I'll fix that.

PR #420, opened 2019-10-25
MHP-3003: Add to Homescreen

OzzieOrca:


OzzieOrca:
I think Travis is failing because of the GraphQL changes. Guess I'll wait till you get your PRs merged.

webpack.config.js
```diff
@@ -88,6 +89,11 @@ module.exports = (env = {}) => {
                                   'src/.well-known/apple-app-site-association',
                           },
                       ]),
+                      new WorkboxPlugin.GenerateSW({
+                          clientsClaim: true,
+                          skipWaiting: true,
+                          navigateFallback: '/index.html',
+                      }),
```
TheNoodleMoose: So does Workbox basically turn the web app into a PWA that can be used offline? That is super cool!

.eslintrc.json
```diff
@@ -89,6 +89,7 @@
 
         "angular/definedundefined": "off",
         "angular/typecheck-string": "off",
-        "angular/no-service-method": "off"
+        "angular/no-service-method": "off",
+        "angular/window-service": "off"
```
TheNoodleMoose: Does this just tell eslint to stop yelling at us for not using angulars window service(window vs $window)?

webpack.config.js
```diff
@@ -76,9 +76,14 @@ module.exports = (env = {}) => {
                       new InlineManifestWebpackPlugin({
                           name: 'webpackManifest',
                       }),
-                      new FaviconsWebpackPlugin(
-                          './app/assets/images/favicon.png',
-                      ),
+                      new FaviconsWebpackPlugin({
```
TheNoodleMoose: Does this create the icon for the PWA version?

webpack.config.js
OzzieOrca: Yep :) It adds the service worker needed for the homescreen prompt on Chrome and does add offline and precaching support :)

.eslintrc.json
OzzieOrca: Ya. I got tired of it. I could just ignore it in individual files if you think we should keep it.

webpack.config.js
OzzieOrca: So it's always been around for favicons. I gave it a higher resolution icon and added some metadata that it will put in manifest.json for us.

TheNoodleMoose: Ah gotcha, cool!

.eslintrc.json
TheNoodleMoose: No, it can be annoying let's just leave it off 👍

PR #422, opened 2019-10-29
MHP-3002: Suggest landscape

OzzieOrca:


app/assets/javascripts/angular/components/myPeopleDashboard/myPeopleDashboard.service.js
```diff
@@ -4,6 +4,11 @@ angular
 
 function myPeopleDashboardService(httpProxy, modelsService, _) {
     var myPeopleDashboardService = {
+        isMobile: () =>
+            window.navigator.userAgent.match(
+                /Android|webOS|iPhone|iPad|iPod|BlackBerry|Windows Phone/i,
```
OzzieOrca: It's back from the dead haha 💀

TheNoodleMoose: Oh no, Halloween came early 😆

TheNoodleMoose:
Should we add the test back in as well? 😆

TheNoodleMoose:
Oh yeah that's true. Yeah its not a very complex test so I think we'll be okay without it haha.

PR #426, opened 2019-11-07
MHP-3035-Update-Challenges-Pagination-Logic-Queries

TheNoodleMoose:


src/communityInsights/components/Table/index.tsx
```diff
@@ -162,6 +162,8 @@ const Table = ({ query, headers, mapRows, mapPage, variables }: Props) => {
     const previousPage = () => {
         if (page.hasPreviousPage) {
             refetch({
+                first: null,
+                last: 5,
```
OzzieOrca:
```suggestion
                last: PAGE_SIZE,
```

src/communityInsights/components/Table/index.tsx
```diff
@@ -162,6 +162,8 @@ const Table = ({ query, headers, mapRows, mapPage, variables }: Props) => {
     const previousPage = () => {
         if (page.hasPreviousPage) {
             refetch({
+                first: null,
```
OzzieOrca: Do you really need `first` and `after` if you're just setting them to `null`/`undefined`?

TheNoodleMoose: Yeah because when you refetch it uses the same variables and those variables values unless you specify different. So if you don't set first to null, it'll fetch both first and last five and things just get broken haha. I also had to update it for the next page refetch as well because if I paginated backwards and forwards it was still trying to grab the last five as well.

src/communityInsights/components/Table/index.tsx
```diff
@@ -152,6 +152,8 @@ const Table = ({ query, headers, mapRows, mapPage, variables }: Props) => {
     const nextPage = () => {
         if (page.hasNextPage) {
             refetch({
+                first: PAGE_SIZE,
+                last: undefined,
```
TheNoodleMoose: Like I said if I paginated backwards and then forwards, the refetch just calls the last query used, with its variables and their values. So last was set to 5 every time which eventually broke things.

src/communityInsights/components/Table/index.tsx
OzzieOrca: Ah ok makes sense.

PR #450, opened 2019-11-26
MHP-1878: adding default values to user notification settings

dbenton9:


app/assets/javascripts/angular/components/user-preferences/user-preferences.component.js
```diff
@@ -10,8 +10,15 @@ class UserPreferences {
...
-            user.notification_settings || {};
+        const {
+            person_moved,
+            person_assigned,
+            weekly_digest,
+        } = user.notification_settings || {
+            person_moved: true,
+            person_assigned: true,
+            weekly_digest: true,
+        };
```
TheNoodleMoose: @OzzieOrca is ```user.notification_settings``` always guaranteed to at least be an empty object? If so you could probably simplify this with:
```suggestion
const {
    person_moved = true,
    person_assigned = true,
    weekly_digest = true,
} = user.notification_settings;
```

dbenton9: @TheNoodleMoose I will default to your and @OzzieOrca decision on this one. One question I have, if we set it up like your suggesting would that automatically overwrite the user attempting to change their notification settings and always have it set to "true"?

TheNoodleMoose: @dbenton9 It won't overwrite the users option, it basically says if this variable doesn't have a value, assign a default value of true to that item.

OzzieOrca: `user.notification_settings` I think starts as `null`. So to combine both of your solutions:
```js
const {
  person_moved = true,
  person_assigned = true,
  weekly_digest = true,
} = user.notification_settings || {};
```

I think Shelby's solution is fine. I think the client defines all 3 fields on the object after it is modified once. But defaulting all of them is probably safer.

dbenton9: @OzzieOrca, just to clarify is this the combined idea you were thinking of?
        const {
            person_moved = true,
            person_assigned = true,
            weekly_digest = true,
        } = user.notification_settings || {
            person_moved: true,
            person_assigned: true,
            weekly_digest: true,
        };

OzzieOrca: So we need to cover the cases where `user.notification_setting` is `null` and where it may not have one of those keys set. (Not sure it will ever end up in that state but I suppose it could if there was a bug or someone edited the DB or sent a rogue API request.)

` || {}` makes sure you have an object you can [destructure](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) with `const { ... } = `, otherwise JS will throw an error about destructuring null/undefined.

The ` = true` will set the value of that key to `true` when the key is `undefined` or doesn't exist. So ` || {}` should be enough and you don't need to set every key there since the default values in the destructuring statement will handle it.

So I think the code in my last comment is enough. It adds the ` || {}` to the code Christian suggested.

dbenton9: Okay, that's starting to make sense. Thank you for taking the time to explain. It's been almost two years since I learned about constructors/deconstructors in school.

OzzieOrca: So these aren't class constructors/destructors which are used to create new objects and clean up after them. This is "destructuring assignment" or "destructuring" which pull parts of an object  or array out:
>The destructuring assignment syntax is a JavaScript expression that makes it possible to unpack values from arrays, or properties from objects, into distinct variables.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

Guess that is kinda confusing...

PR #464, opened 2020-01-14
Fix_React_GraphQL_Mocks

TheNoodleMoose:


src/communityInsights/components/Table/__tests__/Table.tsx
```diff
@@ -25,34 +25,33 @@ describe('<Table />', () => {
         unmount();
     });
 
-    it('should render properly with data', async () => {
+    fit('should render properly with data', async () => {
```
OzzieOrca: Don't commit `fit` :) I think there may be an eslint rule for this...

src/communityInsights/components/Table/__tests__/Table.tsx
```diff
@@ -25,34 +25,33 @@ describe('<Table />', () => {
...
+                    orgId: '1',
+                },
                 mocks: {
-                    Query: () => ({
-                        globalCommunityChallenges: () => ({
-                            nodes: () => [],
-                            pageInfo: () => {},
+                    GlobalCommunity: () => ({
+                        communityChallenges: () => ({
+                            nodes: () => new MockList(2),
```
OzzieOrca: I think it returns a list of 2 items by default...
https://www.apollographql.com/docs/graphql-tools/mocking/#using-mocklist-in-resolvers
>By default, the summary field will always return 2 entries.

src/communityInsights/containers/ChallengesPage/__tests__/ChallengesPage.tsx
```diff
@@ -15,5 +17,24 @@ describe('<ChallengesPage />', () => {
...
 
-    // TODO: Add a test with mocking data when GraphQL schema contains above query
+    fit('should render properly after loading', async () => {
+        const { snapshot } = renderWithContext(<ChallengesPage />, {
+            appContext: {
+                orgId: '1',
+            },
+            mocks: {
+                Community: () => {
+                    return {
```
OzzieOrca: Use the arrow function's default return value?

src/communityInsights/components/Table/__tests__/Table.tsx
TheNoodleMoose: Ah good catch! Will remove. Will it make that test the only test that is ran if it is committed?

src/communityInsights/containers/ChallengesPage/__tests__/ChallengesPage.tsx
TheNoodleMoose: Ah yep, will change!

src/communityInsights/components/Table/__tests__/Table.tsx
OzzieOrca: Ya the only test in that file.

src/testUtils/globalMocks.ts
```diff
@@ -14,14 +14,16 @@ export const globalMocks = {
     Float: () => faker.random.number({ precision: 0.01 }),
     Boolean: () => faker.random.boolean(),
     ID: () => nextId(),
+    ISO8601DateTime: () => faker.date.past(10, '2020-01-14').toUTCString(),
+    ISO8601Date: () => faker.date.past(10, '2020-01-14').toUTCString(),
```
OzzieOrca: I think we want this one just to be the date. I was trying to see if we could do it without moment but the answers seemed more complicated.
```suggestion
    ISO8601Date: () => moment(faker.date.past(10, '2020-01-14')).format('YYYY-MM-DD'),
```

TheNoodleMoose:
@OzzieOrca This should be fine to merge in, right?

OzzieOrca:
Yep :)

PR #466, opened 2020-01-15
Update Helpscout Beacon

OzzieOrca:
(the old beacon is being retired soon)

### DIFF
--- a/index.ejs
+++ b/index.ejs
@@ -52,40 +52,32 @@
                 window.location.origin !== 'https://stage.mhub.cc' &&
                 window.location.origin !== 'https://mhub.cc'
             ) {
-                !(function(e, o, n) {
-                    (window.HSCW = o),
-                        (window.HS = n),
-                        (n.beacon = n.beacon || {});
-                    var t = n.beacon;
-                    (t.userConfig = {}),
-                        (t.readyQueue = []),
-                        (t.config = function(e) {
-                            this.userConfig = e;
+                !(function(e, t, n) {
+                    function a() {
+                        var e = t.getElementsByTagName('script')[0],
+                            n = t.createElement('script');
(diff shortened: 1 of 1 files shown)


index.ejs
```diff
@@ -52,40 +52,32 @@
...
-                        c = e.createElement('script');
-                    (c.type = 'text/javascript'),
-                        (c.async = !0),
-                        (c.src = 'https://djtflbt20bdde.cloudfront.net/'),
-                        r.parentNode.insertBefore(c, r);
-                })(document, window.HSCW || {}, window.HS || {});
-
-                HS.beacon.config({
-                    topArticles: true,
-                });
```
mattdrees: This config wasn't part of the old boilerplate code snippet, so it must have been added specifically. I am not sure what it did, but you might double check that something comparable is not needed with the new beacon.

mattdrees:
Can you guys ping me when this is live, and I'll delete the old HS beacon?

index.ejs
OzzieOrca: I assume that's sorta equivalent to this
Screen Shot 2020-01-16 at 1 51 04 PM
Seems like it's showing mostly the same content as the old one. I don't think it's needed...

OzzieOrca:
@mattdrees I think we are at least waiting on a color change here to `#007398`. Idk what else Spencer might have requested in the email thread.

mattdrees:
I thought I had made that change already, and let Spencer know in slack

soberstadt:
Yeah, I confirmed on the netlify preview the color is right.

OzzieOrca:
Didn't realize the changes were server driven and was waiting for new code or notification that the changes were made. Sorry. Deployed. We can delete the old beacon.

mattdrees:
Great, I deleted it.

PR #471, opened 2020-01-30
MHP-3050: Make survey question editor responsive

OzzieOrca:
Yuck 😰🤢 I feel like rewriting the whole thing in React would have been faster... AngularJS and mutating objects and then relying on references is so buggy.


app/assets/javascripts/angular/components/surveyOverview/surveyOverview.html
```diff
@@ -3,37 +3,37 @@
...
+                ng-class="{'active-tab': $ctrl.activeTab === tab.id}"
+                ng-click="$ctrl.activeTab = tab.id"
+            >
+                <span class="pivot-white">{{ tab.name | t }}</span>
             </div>
         </div>
-        <a
-            ui-sref="appWithoutMenus.previewSurvey({ surveyId: $ctrl.survey.id })"
-            target="_blank"
-            class="self-center f6 link ba ph4 pv3 mr4 white bold br2 dim ttu tracked"
```
TheNoodleMoose: So many classes 😆

app/assets/javascripts/angular/components/surveyOverview/surveyOverview.html
```diff
@@ -3,37 +3,37 @@
...
+                <div class="mr2" ng-include="$ctrl.chevronLeftIcon"></div>
+            </a>
+            <span class="flex-auto">
                 {{ $ctrl.survey.title }}
-            </div>
+            </span>
+            <a
+                ui-sref="appWithoutMenus.previewSurvey({ surveyId: $ctrl.survey.id })"
+                target="_blank"
+                class="self-center f6 link ba ph2 pv1 ph3-ns pv2-ns white bold br2 dim ttu tracked"
```
TheNoodleMoose: So I noticed when testing this locally that when not mobile, the preview link on hover goes completely blue and unseeable. I think something with the tachyon link class here is overriding the white on hover. Changing the white to our pivot-white fixes it to what it used to be.
```suggestion
                class="self-center f6 link ba ph2 pv1 ph3-ns pv2-ns pivot-white bold br2 dim ttu tracked"
```

app/assets/javascripts/angular/components/assignedPeopleSelect/assignedPeopleSelect.service.js
```diff
@@ -11,12 +11,16 @@ function assignedPeopleSelectService(
...
         // Search for people in a particular organization that match the query
         searchPeople: function(query, organizationId, deduper) {
+            if (!query || query.length < 2) {
+                return Promise.resolve([]);
+            }
+
             return httpProxy
                 .get(
-                    modelsService.getModelMetadata('person').url.search,
+                    modelsService.getModelMetadata('person').url.all,
```
TheNoodleMoose: Even when changing this to .all, we are still only searching for people in the current organization due to the `organizations_ids `filter, right? Was there a reason we were using the search endpoint this whole time?

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
```diff
@@ -276,7 +281,12 @@ function surveyOverviewQuestionsController(
     };
 
     this.addPersonToRule = async (question, rule) => {
-        const index = question.question_rules.indexOf(rule);
+        const index = question.question_rules.findIndex(
+            currentRule => currentRule.id === rule.id,
+        );
+
+        // This is terrible. Somehow the reference isn't preserved when editing multiple times.
+        question.question_rules[index] = rule;
```
TheNoodleMoose: Was it always like this, even when I had made the change initially? Or is it something with the new changes that has messed with referencing. AngularJS makes me sad.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/autoAssignLabelNotifyModal.html
```diff
@@ -0,0 +1,70 @@
...
+            <label class="db f6 lh-copy b mb2 pivot-gray-darker ttu tracked">
+                {{ 'surveys:questions.autoAssign' | t }}
+            </label>
+            <assigned-people-select
+                class="db mb2"
+                action-after-select="$ctrl.addPersonToRule($ctrl.question, $ctrl.assignRule)"
+                assigned="$ctrl.assignRule.assign_to"
+                rule-code="$ctrl.assignRule.rule_code"
+                organization-id="$ctrl.survey.organization_id"
+                disabled="$ctrl.assignRule.trigger_keywords === ''"
```
TheNoodleMoose: So I noticed currently on the app, if the keyword/answer is empty we disable these from being used. But with this change if it's empty the user can still click the link to open the modal. But on mobile, the input fields have no indication that they're disabled. Should we instead make the link to open the modal be disabled if the keyword or answer is empty?  That way the user on mobile isn't confused when they open the modal and then can't edit anything. 
Screen Shot 2020-01-31 at 11 50 26 AM

app/assets/javascripts/angular/components/assignedPeopleSelect/assignedPeopleSelect.service.js
OzzieOrca: Historical reasons. The only way to search used to be the `/search` endpoint. Slowly all the functionality got added to `/people`. After this, the API may be able to delete the `/search` endpoint.

app/assets/javascripts/angular/components/surveyOverviewQuestionsTab/surveyOverviewQuestions.component.js
OzzieOrca: Idk. I could debug it more. Maybe it's getting a new object from the API that overwrites the reference... I kinda gave up.

PR #508, opened 2020-02-19
During Travis deploy, lower cache time for .well-known files since th…

OzzieOrca:
…ey don't have unique urls


OzzieOrca:
Ya the default travis deploy config doesn't support different cache times. I haven't looked to see if they've updated it. I agree this is kinda messy. Oh well 🤷‍♂

PR #654, opened 2020-06-25
Temporary marketing landing page for campus contacts split

OzzieOrca:


OzzieOrca:
I redid this on you so we'd still have a way to sign in for testing. Sorry for the premature review. Now the main sign in route should point to the landing page and a secret one `/sign-in-test` should point to the original component.

PR #656, opened 2020-06-25
Add new Facebook App ID for campus contacts

OzzieOrca:
- Change dev environment to hit stage campus contacts api
- Upgrade FB SDK
- Change HelpScout Beacon ID
- Update The Key client ids

### DIFF
--- a/app/assets/javascripts/angular/missionhubApp.config.js
+++ b/app/assets/javascripts/angular/missionhubApp.config.js
@@ -46,7 +46,7 @@ angular
                 },
                 stagingCampusContacts: {
                     apiUrl: 'https://campus-contacts-api-stage.cru.org/apis/v4',
-                    theKeyClientId: '8138475243408077361',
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/angular/missionhubApp.config.js
```diff
@@ -20,50 +20,55 @@ angular
                     'new.missionhub.com',
                     '*.netlify.com',
                     '*.netlify.app',
+                    '*.d11caubtn1mk5o.amplifyapp.com',
```
soberstadt, line 23: is this going to be consistent?

OzzieOrca: Ya it's the Amplify app id. It creates subdomains off of it like pr-656.d11caubtn1mk5o.amplifyapp.com. I changed them around cuz I had to split into multiple Amplify apps to make subdomain specific redirects work.

PR #683, opened 2020-08-25
CC-11: Fix mass edit surveyId

OzzieOrca:


OzzieOrca:
That's how I feel about all this code. It's so abstracted and untyped. You just get the general feeling of "I think this does what I want?". (I did test it.)

PR #699, opened 2020-09-18
Fix survey responses load more

OzzieOrca:
I spent most of my afternoon trying to figure this out and I really have no idea what's happening. I feel like it's some AngularJS internal bug that is dependent on how the function gets triggered.

I haven't fixed this but added a workaround of allowing the user to click the "Loading more..." (and also fixed some styling on the suborgs page).

https://github.com/sroze/ngInfiniteScroll/blob/master/src/infinite-scroll.js#L92 calls the `loadPersonPage` function which runs https://github.com/CruGlobal/missionhub-web/blob/77b4af27ab5eacd0722c4b3860fc1ee3685d2740/app/assets/javascripts/angular/components/peopleScreen/peopleScreen.component.js#L216 every single time but the Angular view never updates. I've tried all the normal AngularJS hacks I can think of and can't get anything to work. If I call `loadPersonPage` in a `ng-click` it works great :(

Also it's only a bug if you get to https://localhost:8080/ministries/16233/survey/19237/responses from another page. If you refresh the page while there, it works great :(

### DIFF
--- a/app/assets/javascripts/angular/components/peopleScreen/peopleScreen.html
+++ b/app/assets/javascripts/angular/components/peopleScreen/peopleScreen.html
@@ -232,7 +232,11 @@
                             show-last-survey="$ctrl.surveyId"
                         >
                         </ministry-view-person>
-                        <div ng-if="!$ctrl.loadedAll" class="row message-row">
+                        <div
+                            ng-if="!$ctrl.loadedAll"
+                            class="row message-row"
+                            ng-click="$ctrl.loadPersonPage()"
+                        >
                             {{ 'general.loading_more' | t }}
                         </div>
                     </div>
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/angular/components/peopleScreen/peopleScreen.html
```diff
@@ -232,7 +232,11 @@
                             show-last-survey="$ctrl.surveyId"
                         >
                         </ministry-view-person>
-                        <div ng-if="!$ctrl.loadedAll" class="row message-row">
+                        <div
+                            ng-if="!$ctrl.loadedAll"
+                            class="row message-row"
```
soberstadt, line 237: Could we underline the text to signal that it's clickable?

soberstadt (changes requested):

soberstadt (approved):

PR #702, opened 2020-09-22
Fix includes on people tab for person page

OzzieOrca:
- Fixes https://secure.helpscout.net/conversation/1270502642/482727/
- Merge params and extra config params to prevent them from overwriting each other in routesService.getPerson
- Include answer_sheets.survey on person history request


soberstadt (approved):
TIL about the `...`

PR #854, opened 2022-06-02
CC-2 - Okta Login

jbirdjavi:
I have it working, but not fully tested nor is it very pretty code. Needs some FE developer help.

Mostly stolen from Scotty's demo work here: https://codesandbox.io/s/missionhub-okta-test-oknqp?file=/src/App.tsx


app/assets/javascripts/angular/campusContactsApp.config.js
```diff
@@ -14,9 +14,10 @@ angular
             },
             vars: {
                 development: {
-                    apiUrl: 'https://campus-contacts-api-stage.cru.org/apis/v4',
+                    apiUrl: 'http://localhost:3000/apis/v4',
```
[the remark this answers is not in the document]
jbirdjavi: Oops, I didn't mean to commit that file. But no, it didn't redirect to the staging site for my local testing. I put `https://localhost:8080` as a redirect_uri in the Okta application settings, so it should work fine.

wrandall22:
@jbirdjavi is the API code already in stage for testing against?

jbirdjavi:
@wrandall22 Yes, but it didn't have the correct secrets settings until just now.

app/assets/javascripts/angular/services/oktaService.js
```diff
@@ -0,0 +1,53 @@
+import { OktaAuth } from "@okta/okta-auth-js";
+
+angular.module('campusContactsApp').factory('oktaService', oktaService);
+
+const authClient = new OktaAuth({
+    issuer: "https://signon.okta.com/oauth2/default",
+    clientId: "redacted",
```
wrandall22, line 7: Not much point making this redacted. If it's in the code here it will be visible to anyone looking for it. If this is a sensitive value it can't be here in this JS code.

wrandall22: If we are using the same app for stage and prod we can probably hard code it here. If we are using different ones we can make an environment variable instead.

jbirdjavi: It should use the same app for both stage and prod. You're definitely right on the "redacted" part. I wasn't sure if it should be set here, or in the other file that has the key and facebook ids.

wrandall22:
For some reason it's trying to redirect me to the admin console: `https://signon-admin.okta.com/admin/app/oidc_client/instance/0oa1oqxpke9BXTbhp0h8#tab-general`. Maybe I don't have the right ids?

wrandall22:
This is on staging now @jbirdjavi do you know what the QA process is?

jbirdjavi:
I don't really. Brian Funkhouser is product for CC but he's not available much this summer. Maybe Nikki Neary has an idea?

My thoughts on things to QA (feel free to add):
- Can log in an existing account
- Can create a new account and login
- Can still log in via FB if that was already active before creating new account
- Logout (logs out from Okta too?)

wrandall22:
The only one that doesn't seem to work well (in stage at least) is 
> Can still log in via FB if that was already active before creating new account

It seems the Facebook application is not set up correctly to handle the stage site.

jbirdjavi:
@wrandall22 I found the settings in the FB app so FB login is now working on staging.

wrandall22 (approved):

PR #867, opened 2023-01-30
Travis -> GitHub Actions Conversion

j2trumpet:


wrandall22:
You'll need to delete the travis file, and should squash all these commits together.

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,72 @@
...
+              with:
+                  node-version: '12'
+
+            - name: 📦 Install dependencies
+              run: yarn install
+
+            - name: Curl connection
+              uses: wei/curl@v1
+              with:
+                  args: -s https://codecov.io/bash
```
wrandall22, line 66: Looks like this is [deprecated](https://docs.codecov.com/docs/about-the-codecov-bash-uploader)

j2trumpet: I've swapped this out for the [official codecov action](https://github.com/marketplace/actions/codecov). The action passes, though it says it didn't find any coverage files. There weren't any files specified in the travis script and I can't see if there was anything specified in travis itself.

dr-bizz (changes requested):
Great job. I see a few tweaks but not much.

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,70 @@
+name: OneSky Upload
+
+on:
+    push:
+        branches:
+            - travis-gha-conversion
```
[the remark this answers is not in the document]
j2trumpet: @dr-bizz So I should target both `master` and `staging`? The Travis file is identical between the two branches.

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,70 @@
...
+    pull_request:
+        branches:
+            - travis-gha-conversion
+    workflow_dispatch:
```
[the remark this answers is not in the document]
j2trumpet: @dr-bizz If I understand `workflow_dispatch` correctly (and it's possible I don't), it allows us to run the workflow manually - no push required. If something happened on OneSky and we needed to reupload the `en-US.js` file, this ability could come in handy.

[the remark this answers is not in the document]
dr-bizz: No, you're correct. I didn't understand what `workflow_dispatch `did, so I assumed it needed some options/inputs for it to run.
Looking through the documentation, I see that you have configured it correctly and the options/inputs are optional. - https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

.github/workflows/node.js.yml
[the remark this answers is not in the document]
dr-bizz: Yes, I believe so. You'll need to configure `push` to have both `master` and `staging` branches and `pull_request` only to have the `master` branch since I don't believe devs do PRs to staging on this repo.

I am a little confused by the travis.yml file. It doesn't mention any branches, so I assume that means Travis CI runs on all branches? https://docs.travis-ci.com/user/customizing-the-build/#building-specific-branches
If that is the case, you'll need to do something like this:
```
on:
  push:
    branches:    
      - '**'
```
Ref: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#on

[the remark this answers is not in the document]
j2trumpet: @dr-bizz I believe this syntax works for targeting every `push` regardless of branch.
```
on:
  push:
  pull_request: ...
```
However, I've noticed since I first added any mention of branches other than my feature branch to `push` and `pull_request` this morning that the script is getting triggered twice and I can't figure out why. `pull_request` should only be firing on `master`.

.github/workflows/node.js.yml
j2trumpet: @wrandall22 Does my update look good?

.github/workflows/node.js.yml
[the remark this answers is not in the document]
dr-bizz: I think because you're pushing to your branch, it triggers the `pull` and then because you're on a PR merging into master, you also trigger the `pull_request`.

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,67 @@
...
+
+            - name: 🏗️ Build app
+              run: yarn build --prod
+
+            - name: Upload coverage to Codecov
+              uses: codecov/codecov-action@v3
+
+    deploy:
+        name: OneSky Upload
+        if: github.event_name != 'pull_request' && github.ref == 'refs/heads/travis-gha-conversion'
```
dr-bizz, line 47:
```suggestion
        if: github.event_name != 'pull_request' && github.ref == 'refs/heads/master'
```

dr-bizz: I'm guessing you left this here so you can test it, but just wanted to double check.

j2trumpet: @dr-bizz Yes, that needs to be done, but I'm waiting to do it until it's the only thing that needs doing so that I can continue to make sure this upload is working with any other changes I make. Is this the only thing left to do before you'll approve it?

.github/workflows/node.js.yml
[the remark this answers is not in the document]
j2trumpet: @dr-bizz Didn't you explain that above? Because I'm pushing to a branch that has an active PR, it's triggering `pull_request` in addition to `push`?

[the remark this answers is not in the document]
j2trumpet: @dr-bizz I think I may have had a syntax error, though I would have expected GHA to fail if that were the case. It's only running only once now. When I first pushed the syntax change, it ran twice, but it's run only once on the two pushes since then.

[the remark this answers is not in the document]
dr-bizz: Brilliant!

dr-bizz:
@j2trumpet, code looks good to me! AWS Amplify is failing; upon investigating, the server worked when the pull_request branches were set to `travis-gha-conversion` instead of `- master`.

.github/workflows/node.js.yml
wrandall22: You may need to pass in the token, not sure. I saw someone do that recently.

wrandall22 (approved):

.github/workflows/node.js.yml
j2trumpet: @wrandall22 The [documentation](https://github.com/marketplace/actions/codecov#usage) says a token isn't required for public repos and this one is public.

j2trumpet:
@wrandall22 I've removed the Travis file, but the PR is still looking for it. I see a [webhook](https://github.com/CruGlobal/campus-contacts-web/settings/hooks) pointing to `https://notify.travis-ci.org/` and Travis CI listed in the repo's [app integrations](https://github.com/CruGlobal/campus-contacts-web/settings/installations) (though I can't configure it). What do I need to do to fully remove Travis?

wrandall22:
It is on the branch protection settings for the `master` branch.

j2trumpet:
@wrandall22 So am I clear to get rid of that protection setting?

wrandall22:
Yes, specifically the one saying that Travis needs to succeed (not the whole branch protection).

j2trumpet:
@dr-bizz Have I made the changes you requested? If so, can you mark your suggestion as completed?

dr-bizz (approved):

.github/workflows/node.js.yml
wrandall22: @j2trumpet you forgot to update this before merge.

wrandall22: You need a new PR

PR #870, opened 2023-10-11
Upgrading GHA to Node 16

dr-bizz:
## Description
Upgrading GHA to use Node 16, to ensure everything is working on Node 16 before deployment.

## Changes
- Switched from using node 12 to node 16 on GHA

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -929,9 +929,9 @@
     to-fast-properties "^2.0.0"
 
 "@brainly/onesky-utils@^1.4.1":
-  version "1.4.1"
-  resolved "https://registry.yarnpkg.com/@brainly/onesky-utils/-/onesky-utils-1.4.1.tgz#ee2d15567e9c617b534295fdd50abea9a3a7593e"
-  integrity sha512-Vi4u1xGUxS0IKqNZkHsgrd9z53/6P3oqdCV6Ao7O9itScnYjXfy00tZoagmAbrLtqbwsI4O08HjVQ2aYwW5vSg==
+  version "1.4.2"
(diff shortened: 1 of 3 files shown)


wrandall22 (approved):
Wow, did we really need no other updates?

dr-bizz:
They are coming in another PR

yarn.lock
```diff
@@ -929,9 +929,9 @@
     to-fast-properties "^2.0.0"
 
 "@brainly/onesky-utils@^1.4.1":
-  version "1.4.1"
-  resolved "https://registry.yarnpkg.com/@brainly/onesky-utils/-/onesky-utils-1.4.1.tgz#ee2d15567e9c617b534295fdd50abea9a3a7593e"
-  integrity sha512-Vi4u1xGUxS0IKqNZkHsgrd9z53/6P3oqdCV6Ao7O9itScnYjXfy00tZoagmAbrLtqbwsI4O08HjVQ2aYwW5vSg==
+  version "1.4.2"
```
wrandall22, line 932: No change needed in package.json?

dr-bizz: I updated the package `@brainly/onesky-utils` from `1.4.1` to `1.4.2`, so there wasn't any need for a package.json edit.

I updated `yarn.lock` by running `yarn upgrade @brainly/onesky-utils`.

wrandall22: How, then, will it handle [this line](https://github.com/CruGlobal/campus-contacts-web/blob/master/package.json#L62)?

dr-bizz: I can update it there too and will do for clarity. But since we have the version set to `^1.4.1`. It is going to install both minor and patch updates. which is why it installs version `1.4.2`.

wrandall22: Ah, I see. I missed the `^`.

wrandall22 (approved):

PR #873, opened 2024-02-26
Upgrade to node v18

canac:
## Description

- Upgrade to node v18
- Manually merged into staging and tested in staging ✅
- Before merging into production, the `_CUSTOM_IMAGE` Amplify environment variable should be set to `public.ecr.aws/docker/library/node:18` on all branches.

## Changes
- Add node `--openssl-legacy-provider` flag to `package.json` scripts until we can upgrade to webpack v5 and upgrade karma
- Upgrade `postcss` subdependency

## Checklist:
- [x] I have given my PR a title with the format "CC-(JIRA#) (summary sentence max 80 chars)"
- [ ] I have applied the appropriate labels
- [x] I have requested a review from another person on the project


wrandall22 (approved):
Stage appears to be working well

dr-bizz:
@canac Using `NODE_OPTIONS=--openssl-legacy-provider` means we're using an insecure algorithm for OpenSSL.

> The OpenSSL legacy provider supplies OpenSSL implementations of algorithms that have been deemed legacy. Such algorithms have commonly fallen out of use, have been deemed insecure by the cryptography community, or something similar.
> 
> We can consider this the retirement home of cryptographic algorithms.

[https://www.openssl.org/docs/man3.0/man7/OSSL_PROVIDER-legacy.html](https://www.openssl.org/docs/man3.0/man7/OSSL_PROVIDER-legacy.html#:~:text=The%20OpenSSL%20legacy%20provider%20supplies,retirement%20home%20of%20cryptographic%20algorithms.)

It's best we don't use `--openssl-legacy-provider` but instead upgrade Webpack. (I think that's how I got around it on MPDx_web)

canac:
@dr-bizz I agree that that's not ideal long-term. However, I believe that Bill said that this tool is going away soon. And it is just the build process accepting legacy OpenSSL algorithms, not the actual server, so my understanding is that the security risk is minimal. Also, my understanding is that it's not any less secure than it was before, this flag just opts us back into the node v16 behavior.

@wrandall22 Is upgrading the whole tech stack (specifically webpack v4 -> v5) in scope for the OKR task, or did we just want to do the minimum necessary work to make sure it still runs on Amplify?

wrandall22:
I think for this tool the way we did it here is fine.

PR #874, opened 2025-04-09
[CC-15] Add sunset banner

canac:
## Description

Add a sunset banner to Campus Contacts. The names of the user's organizations are easy to pull from the API. I could not figure out how to extract more information because the API doesn't expose the Infobase id `importable_id` field that is in the database.

I didn't make the text into a translation because this app is going away.

## Changes

[CC-15](https://jira.cru.org/browse/CC-15)

## Checklist:
- [x] I have given my PR a title with the format "CC-(JIRA#) (summary sentence max 80 chars)"
- [x] I have applied the appropriate labels
- [x] I have requested a review from another person on the project

### DIFF
--- a/app/components/navigation/navHeader.component.js
+++ b/app/components/navigation/navHeader.component.js
@@ -24,4 +24,8 @@ function navHeaderController(state, loggedInPerson, envService, authenticationSe
   };
 
   this.toggleSearchBar = () => (this.showSearchBar = !this.showSearchBar);
+
+  this.helpEmailBody = encodeURIComponent(`Team Name (from Infobase):
+Team Leader email:
+Possible Team Admin:`);
 }
--- a/app/components/navigation/navHeader.html
+++ b/app/components/navigation/navHeader.html
@@ -1,3 +1,13 @@
+<div class="sunset-banner">
+  <h2>
+    Attention: Campus Contacts will be retired on May 31, 2025. Learn about our replacement tool here:
+    <a href="https://docs.google.com/document/d/11v9feEK3EIP3AtCzDV8Xkg8v8sYbdL2EHQylqH4TK-Q/edit" target="_blank"
+      >Connections[Beta] FAQ</a
+    >. If you need assistance transitioning to the new tool, please reach out to your team leader or
+    <a ng-href="mailto:help@cru.org?subject=Onboarding+to+Connections&body={{$ctrl.helpEmailBody}}">help@cru.org</a>
+    for support.
+  </h2>
+</div>
 <div class="main-menu bg-pivot-dark-blue flex items-center justify-between ph2 ph5-ns">
   <ul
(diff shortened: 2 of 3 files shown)


app/components/navigation/navHeader.component.js
```diff
@@ -24,4 +31,16 @@ function navHeaderController(state, loggedInPerson, envService, authenticationSe
   };
 
   this.toggleSearchBar = () => (this.showSearchBar = !this.showSearchBar);
+
+  this.helpEmailBody = () => {
+    const body = `Team(s) from Infobase:${this.orgs.map((org) => `\n* ${org.name}`).join('')}`;
```
wrandall22, line 36: In theory, we should be able to get the possible admin data as we are kind of displaying it in the UI. E.g. if user's name exists in team and that contact has admin privileges. Here is my record on one of my teams:

Screenshot 2025-04-09 at 1 16 12 PM

canac: I'm not very familiar with this tool, so thanks for pointing this out. Is "Possible Admins" supposed to look through each team's contacts to show who is admins on each team? Or to show which teams the logged in user administrates? From your comment, it sounds like the latter.

wrandall22: I think the goal is closer to the latter, but probably broader in nature. If the logged in user administrates any teams, we would pass true or something. I can get more clarity, but that would make sense to me as it seems to be trying to indicate what level of license they will need in Salesforce.

wrandall22: Just got more clarity. The idea is that the user will fill in the Team Leader and the Possible Team Admin data manually as the email body comes up. Here's an example:

wrandall22: Looks like we want that to be true for the Team Name as well since the teams in Campus Contacts do not necessarily exist or match up with teams in Infobase.

canac:
The `yarn onesky:download` step during the staging deployment in Amplify is failing with an authentication error. I'm trying to get new OneSky API keys to see if that helps.

app/components/navigation/navHeader.html
```diff
@@ -1,3 +1,13 @@
+<div class="sunset-banner">
+  <h2>
+    Attention: Campus Contacts will be retired on May 31, 2025. Learn about our replacement tool here:
+    <a href="https://docs.google.com/document/d/11v9feEK3EIP3AtCzDV8Xkg8v8sYbdL2EHQylqH4TK-Q/edit"
```
wrandall22, line 4: Should we make this open in a new tab?

wrandall22 (approved):
Everything looks pretty good to me, one other comment (I prefer new tabs myself).

app/components/navigation/navHeader.html
canac: We definitely should since it links to a different site. Thanks for pointing that out.

